Issue #39, opened 2021-06-01
Add documentation for migrating from Elasticsearch OSS to OpenSearch


aetter:
* Conceptual information on the basic goals here
* Paths from Elasticsearch OSS 5.x onward
* Rolling upgrade
* Cluster restart upgrade
* Kibana to OpenSearch Dashboards

PR #40, opened 2021-06-03
Initial migration content

aetter:
### Description
Migration content for moving from Elasticsearch OSS to OpenSearch RC1.

### Issues Resolved
https://github.com/opensearch-project/documentation-website/issues/39
 
### Check List


docs/migrate/docker-migrate.md
```diff
@@ -0,0 +1,14 @@
+---
+layout: default
+title: Docker migration
+parent: Migrate to OpenSearch
+nav_order: 25
+---
+
+# Docker migration
+
+If you use Docker images, we highly recommend that you perform what amounts to a [cluster restart upgrade](../upgrade-migrate/). Update your cluster configuration with new images, new settings, and new environment variables. Then stop and start the cluster. This process requires downtime, but takes very few steps.
```
dblock, line 10: We should explain why.

docs/migrate/index.md
```diff
@@ -0,0 +1,11 @@
...
+layout: default
+title: Migrate to OpenSearch
+nav_order: 6
+has_children: true
+has_toc: true
+---
+
+# Migrate to OpenSearch
+
+The process of migrating from Elasticsearch OSS (including Open Distro for Elasticsearch) to OpenSearch varies depending on your current version of Elasticsearch OSS, install type, tolerance for downtime, and cost-sensitivity. Rather than concrete steps to cover every situation, we have general guidance for the process.
```
dblock, line 11: I think this is about "upgrading" to OpenSearch, not migrating.

dblock (changes requested):
1. We're saying "migrate" a lot. This is an upgrade, so I would rename that everywhere.
2. It's unclear what we actually recommend and why. I don't think recommending downtime is generally good unless there's substantial risk. We should emphasize that rolling upgrade without downtime is possible and/or recommended, but that you can do it faster (?) as a restart upgrade.

docs/migrate/upgrade-migrate.md
```diff
@@ -0,0 +1,176 @@
...
+1. Wait for the node to rejoin the cluster (rolling) or for the cluster to start (cluster restart). Check the `_nodes` summary to verify that all nodes are available and running the expected version:
+
+   ```bash
+   # Elasticsearch OSS
+   curl -XGET 'localhost:9200/_nodes/_all?pretty=true'
+   # Open Distro for Elasticsearch
+   curl -XGET 'https://localhost:9200/_nodes/_all?pretty=true' -u 'admin:admin' -k
+   ```
+
+   Specifically, check the `nodes.node-id.version` portion of the response.
```
keithhc2, line 87: Can we make it clear that `node-id` is a variable?

docs/migrate/upgrade-migrate.md
```diff
@@ -0,0 +1,176 @@
...
+1. Wait for the OpenSearch node to rejoin the cluster (rolling) or for the cluster to start (cluster restart). Check the `_nodes` summary to verify that all nodes are available and running the expected version:
+
+   ```bash
+   # Security plugin disabled
+   curl -XGET 'localhost:9200/_nodes/_all?pretty=true'
+   # Security plugin enabled
+   curl -XGET 'https://localhost:9200/_nodes/_all?pretty=true' -u 'admin:admin' -k
+   ```
+
+   Specifically, check the `nodes.node-id.version` portion of the response.
```
keithhc2, line 163: Ditto above comment.

docs/migrate/docker-migrate.md
aetter: Sure, tweaked the Docker stuff a bit.

docs/migrate/index.md
aetter: I don't think so, actually. If we take the user goal as "I have an Elasticsearch OSS cluster and would like to run OpenSearch instead," then there are a few ways to get there. "Upgrading" to OpenSearch is just one of them. Node replacement is another, and snapshot restore is probably the safest path. All approaches ultimately get you to the same place.

And of course, we'll eventually have an "Upgrade OpenSearch" section, so it seems worthwhile to put some distance between the two and not have the top-level be "Upgrade to OpenSearch."

migrate

dblock: I think this is worth a discussion. I feel pretty strongly that we want to call this "upgrading" because there's continuity in OSS versions, and upgrading doesn't require downtime.

Maybe @elfisher wants to take this discussion on?

dblock: "Upgrade migration" sounds strange ;)

aetter: > "Upgrade migration" sounds strange ;)

No objection on that, and I'll change the title for that page. But the top-level is still about getting to OpenSearch, not necessarily upgrading to get there.

PR #54, opened 2021-06-17
Adds both nodes for dashboards to talk to in docker-compose

elfisher:
### Description
Enables OpenSearch Dashboards to talk to both OpenSearch nodes in the event one is down. To reproduce. Run the compose file and terminate node1. Dashboards will still talk to node2 (although the cluster will be yellow).
 
### Related Issues/PRs 
https://github.com/opensearch-project/project-website/pull/137 
 
### Check List

### DIFF
--- a/_opensearch/install/docker.md
+++ b/_opensearch/install/docker.md
@@ -138,7 +138,7 @@ services:
     expose:
       - "5601"
     environment:
-      OPENSEARCH_HOSTS: https://opensearch-node1:9200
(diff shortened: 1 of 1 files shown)


aetter:
I tested this and was pretty shocked to see that startup fails without the `'`s, and if you have a space between the hosts, that makes it fail, too. I'll merge, dupe over to the Docker security page, and add some comments about not messing with the formatting.

PR #62, opened 2021-06-23
Add Java and OS recommendations

aetter:
### Check List

### DIFF
--- a/_opensearch/install/compatibility.md
+++ b/_opensearch/install/compatibility.md
@@ -8,7 +8,9 @@ nav_order: 2
 # Operating system and JVM compatibility
 
 - We recommend installing OpenSearch on RHEL- or Debian-based Linux distributions that use [systemd](https://en.wikipedia.org/wiki/Systemd), such as CentOS, Amazon Linux 2, and Ubuntu (LTS). OpenSearch should work on many Linux distributions, but we only test a handful.
-- The OpenSearch tarball ships with a compatible version of Java in the `jdk` directory. To find its version, run `./bin/opensearch --version`.
+- The OpenSearch tarball ships with a compatible version of Java in the `jdk` directory. To find its version, run `./jdk/bin/java -version`. For example, the OpenSearch 1.0.0 tarball ships with Java 15 (non-LTS).
+
+{% comment %}`./jdk/bin/java -version` doesn't work on macOS with zsh at the moment, and I have no idea why. Maybe we need a macOS artifact. Regardless, the command works on Amazon Linux 2 with bash and presumably other distros. - aetter{% endcomment %}
 
   To use a different Java installation, set the `OPENSEARCH_JAVA_HOME` environment variable to the Java install location. We recommend Java 11 (LTS), but OpenSearch also works with Java 8.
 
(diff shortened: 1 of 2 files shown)


_opensearch/install/docker.md
```diff
@@ -16,7 +16,7 @@ docker pull opensearchproject/opensearch-dashboards:{{site.opensearch_version}}
 
 To check available versions, see [Docker Hub](https://hub.docker.com/u/opensearchproject).
 
-OpenSearch images use `centos:7` as the base image. If you run Docker locally, we recommend allowing Docker to use at least 4 GB of RAM in **Preferences** > **Resources**.
+OpenSearch images use `amazonlinux:2` as the base image. If you run Docker locally, set Docker to use at least 4 GB of RAM in **Preferences** > **Resources**.
```
aetter, line 19: Hey @peterzhuamazon, this change is the result of my sleuthing, but let me know if I got it wrong.

_opensearch/install/compatibility.md
```diff
@@ -0,0 +1,17 @@
...
+parent: Install OpenSearch
+nav_order: 2
+---
+
+# Operating system and JVM compatibility
+
+- We recommend installing OpenSearch on RHEL- or Debian-based Linux distributions that use [systemd](https://en.wikipedia.org/wiki/Systemd), such as CentOS, Amazon Linux 2, and Ubuntu (LTS). OpenSearch should work on many Linux distributions, but we only test a handful.
+- OpenSearch ships with a compatible version of Java in the `jdk` directory. To find its version, run `./bin/opensearch --version`.
+
+  To use a different Java installation, set the `OPENSEARCH_JAVA_HOME` environment variable to the Java install location. We recommend Java 11 (LTS), but OpenSearch also works with Java 8.
```
aetter, line 13: It looks like our gradle build can test on a wide variety of Java versions (8-12?), but my understanding is that we only want people using the LTS ones for self-installs (8, 11, eventually 17). Again, let me know if I've got it wrong here.

_opensearch/install/compatibility.md
```diff
@@ -0,0 +1,17 @@
...
+layout: default
+title: Compatibility
+parent: Install OpenSearch
+nav_order: 2
+---
+
+# Operating system and JVM compatibility
+
+- We recommend installing OpenSearch on RHEL- or Debian-based Linux distributions that use [systemd](https://en.wikipedia.org/wiki/Systemd), such as CentOS, Amazon Linux 2, and Ubuntu (LTS). OpenSearch should work on many Linux distributions, but we only test a handful.
+- The OpenSearch tarball ships with a compatible version of Java in the `jdk` directory. To find its version, run `./bin/opensearch --version`.
```
dblock, line 11: We should say which version it is. Currently I believe it's 15.

_opensearch/install/compatibility.md
```diff
@@ -0,0 +1,17 @@
+---
+layout: default
+title: Compatibility
+parent: Install OpenSearch
+nav_order: 2
+---
+
+# Operating system and JVM compatibility
```
dblock, line 8: I think this should be split into two sections, OS and JVM

_opensearch/install/compatibility.md
aetter: Sure, added.

_opensearch/install/compatibility.md
aetter: This is already such a short page that I don't see a lot of utility in breaking things up further, especially if you think about it as "all the stuff you have to have in place prior to installing."

PR #66, opened 2021-06-25
Added create alias to API reference page

keithhc2:
Added a create alias page to API reference page
 
### Check List

### DIFF
--- a/_opensearch/rest-api/create-alias.md
+++ b/_opensearch/rest-api/create-alias.md
@@ -1,14 +1,15 @@
 ---
 layout: default
-title: Create alias
+title: Alias
 parent: REST API reference
 grand_parent: OpenSearch
 nav_order: 8
 ---
 
-# Create alias
+# Alias
 
-An alias is a virtual pointer that you can use to reference one or more indices. For example, if you have continuous streams of incoming data but only want to search through the past seven days at any time, you can use an alias to point to only indices created within the past seven days.
+An alias is a virtual pointer that you can use to reference one or more indices.
+
 
 ## Example
 
@@ -21,14 +22,27 @@ POST _aliases
         "index": "movies",
         "alias": "movies-alias1"
       }
+    },
+    {
+      "remove": {
+        "index": "old-index",
+        "alias": "old-index-alias"
+      }
     }
+
   ]
 }
 ```
 
+## Path and HTTP Methods
+
+```
+POST _aliases
+```
+
 ## URL parameters
 
-Both create alias parameters are optional.
+All alias parameters are optional.
 
 Parameter | Data Type | Description
 :--- | :--- | :---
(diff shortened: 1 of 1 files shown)


_opensearch/rest-api/create-alias.md
```diff
@@ -0,0 +1,66 @@
+---
+layout: default
+title: Create alias
```
aetter, line 3: Hey @keithhc2, I believe this method + path combination (POST _aliases) is actually the "add or remove aliases" operation, so calling the page "Create alias" isn't quite right.

_opensearch/rest-api/create-alias.md
```diff
@@ -0,0 +1,66 @@
...
+    {
+      "add": {
+        "index": "movies",
+        "alias": "movies-alias1"
+      }
+    }
+  ]
+}
+```
+
```
aetter, line 28: We want to include the "Path and HTTP methods" section on each of these REST API pages (see https://docs-beta.opensearch.org/opensearch/rest-api/cluster-allocation/#path-and-http-methods for example).

I missed flagging that that section was missing on the https://docs-beta.opensearch.org/opensearch/rest-api/create-index/ page, so I'd either add that section to this PR or do a separate commit -- should be an easy fix regardless.

_opensearch/rest-api/create-alias.md
```diff
@@ -0,0 +1,66 @@
...
+layout: default
+title: Create alias
+parent: REST API reference
+grand_parent: OpenSearch
+nav_order: 8
+---
+
+# Create alias
+
+An alias is a virtual pointer that you can use to reference one or more indices. For example, if you have continuous streams of incoming data but only want to search through the past seven days at any time, you can use an alias to point to only indices created within the past seven days.
```
aetter, line 11: The big thing here is that aliases give your clients a consistent path, so they can always just search "logs-this-week/_search" instead of having to figure out the day and provide the correct index names.

_opensearch/rest-api/create-alias.md
```diff
@@ -0,0 +1,66 @@
...
+        "alias": "movies-alias1"
+      }
+    }
+  ]
+}
+```
+
+## URL parameters
+
+Both create alias parameters are optional.
```
aetter, line 31: I'd say "All alias parameters are optional." since a) this isn't just alias creation and b) that way we don't have to change this sentence if we ever add a third parameter.

_opensearch/rest-api/create-alias.md
```diff
@@ -0,0 +1,66 @@
...
+An alias is a virtual pointer that you can use to reference one or more indices. For example, if you have continuous streams of incoming data but only want to search through the past seven days at any time, you can use an alias to point to only indices created within the past seven days.
+
+## Example
+
+```json
+POST _aliases
+{
+  "actions": [
+    {
+      "add": {
```
aetter, line 20: Include a second, remove action in the example?

aetter (commented):
Looks good to me, minor changes.

Issue #71, opened 2021-07-06
Document migrating from Kibana OSS to OpenSearch Dashboards


dblock:
Coming from #39 we need documentation on migrating from Kibana OSS to OpenSearch Dashboards.

PR #74, opened 2021-07-07
Add page for migrating to Dashboards

aetter:
### Description
 
### Issues Resolved
https://github.com/opensearch-project/documentation-website/issues/71
 
### Check List

### DIFF
--- a/_upgrade-to/dashboards-upgrade-to.md
+++ b/_upgrade-to/dashboards-upgrade-to.md
@@ -1,0 +1,36 @@
+---
+layout: default
+title: Upgrade from Kibana OSS to OpenSearch Dashboards
+nav_order: 50
+redirect_from:
+  - /migrate/dashboards/
+---
+
+# Upgrade from Kibana OSS to OpenSearch Dashboards
+
+Kibana OSS stores its visualizations and dashboards in one or more indices (`.kibana*`) on the Elasticsearch OSS cluster. As such, the most important step is to leave those indices intact as you upgrade from Elasticsearch OSS to OpenSearch.
+
+Consider exporting all Kibana objects prior to starting the upgrade. In Kibana, choose **Stack Management**, **Saved Objects**, **Export objects**.
+{: .tip }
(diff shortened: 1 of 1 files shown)


aetter:
@tmarkley @kavilla

dblock:
I don't know about the implementation details to review this, I think @tmarkley or maybe @mihirsoni?

dblock:
@saratvemulapalli

aetter:
I'm very happy to make changes based on reviewer feedback, but I'm merging now so that we have this content for 1.0.0.

kavilla:
Hey thanks for this. Looks pretty good.

> Port your settings from `<kibana-dir>/config/kibana.yml` to `<dashboards-dir>/config/opensearch_dashboards.yml`.

I think to be safe, I would tell people to create a backup of kibana.yml. Just in case they moved their file and rename it.

> In general, settings with `elasticsearch` in their names map to `opensearch` (e.g. `elasticsearch.shardTimeout` and `opensearch.shardTimeout`) and settings with `kibana` in their names map to `opensearchDashboards` (e.g. `kibana.defaultAppId` and `opensearchDashboards.defaultAppId`). Most other settings use the same names.

This is true, thanks for putting that. But subsequent changes were made to allow for the previous settings to still work with a deprecation message. Should we mention as a note that they shouldn't have to rename the settings right away? We thought it was a better experience for people to just have to rename the file and it work out of the box with warning messages.

aetter:
> This is true, thanks for putting that. But subsequent changes were made to allow for the previous settings to still work with a deprecation message. Should we mention as a note that they shouldn't have to rename the settings right away? We thought it was a better experience for people to just have to rename the file and it work out of the box with warning messages.

Got it, thanks for reviewing. I must have tested right before that change made its way in; I tried porting `kibana.yml` over as-is, and Dashboards failed to start. I'll test again later today and modify as necessary. In the meantime, I've updated the docs with a recommendation to take a backup.

PR #99, opened 2021-07-15
Added update mapping to API reference and fixed some metadata

keithhc2:
Added update index mapping to API reference
 
### Check List

### DIFF
--- a/_opensearch/rest-api/update-mapping.md
+++ b/_opensearch/rest-api/update-mapping.md
@@ -1,0 +1,78 @@
+---
+layout: default
+title: Update mapping
+parent: REST API reference
+nav_order: 6
+---
+
+# Update mapping
+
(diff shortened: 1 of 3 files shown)


_opensearch/rest-api/update-mapping.md
```diff
@@ -0,0 +1,78 @@
+---
+layout: default
+title: Update mapping
+parent: REST API reference
+nav_order: 6
+---
+
+# Update mapping
+
+If you want to update an index's mappings after creating the index, you can do so with the update mapping API operation.
```
aetter, line 10: Any implications if I try to do this on an index with a whole bunch of data? Do the updates have to be additive rather than turning existing strings into integers (I assume this sort of request would error out)?

Basically, a bit more context what why I might do this and any gotchas to be aware of.

aetter (approved):

PR #162, opened 2021-08-30
Added count REST API

ashwinkumar12345:
### Description
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List

### DIFF
--- a/_opensearch/rest-api/count.md
+++ b/_opensearch/rest-api/count.md
@@ -1,0 +1,103 @@
+---
+layout: default
+title: Count
+parent: REST API reference
+nav_order: 150
+---
+
+# Count
+Introduced 1.0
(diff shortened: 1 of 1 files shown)


_opensearch/rest-api/count.md
```diff
@@ -0,0 +1,103 @@
...
+{
+  "query": {
+    "term": {
+      "response": "200"
+    }
+  }
+}
+```
+
+You can also use the search API for the same result:
```
aetter, line 31: I think it's worth being explicit here that _count is really just search without the results. This is a great sample call, but can we introduce it with something more like, "The following call to the search API produces equivalent results"?

aetter (approved):

PR #167, opened 2021-09-08
Update generate-certificates.md

ict-one-nl:
By default certificates are valid for 30 days, that's not useful, set it to 10 years.

### DIFF
--- a/_security-plugin/configuration/generate-certificates.md
+++ b/_security-plugin/configuration/generate-certificates.md
@@ -99,26 +99,26 @@ If you already know the certificate details and don't want to specify them inter
 #!/bin/sh
 # Root CA
 openssl genrsa -out root-ca-key.pem 2048
-openssl req -new -x509 -sha256 -key root-ca-key.pem -subj "/C=CA/ST=ONTARIO/L=TORONTO/O=ORG/OU=UNIT/CN=ROOT" -out root-ca.pem -days 730
+openssl req -new -x509 -sha256 -days 3650 -key root-ca-key.pem -subj "/C=CA/ST=ONTARIO/L=TORONTO/O=ORG/OU=UNIT/CN=ROOT" -out root-ca.pem -days 730
 # Admin cert
 openssl genrsa -out admin-key-temp.pem 2048
 openssl pkcs8 -inform PEM -outform PEM -in admin-key-temp.pem -topk8 -nocrypt -v1 PBE-SHA1-3DES -out admin-key.pem
-openssl req -new -key admin-key.pem -subj "/C=CA/ST=ONTARIO/L=TORONTO/O=ORG/OU=UNIT/CN=ADMIN" -out admin.csr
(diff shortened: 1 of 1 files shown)


aetter:
Hey @ict-one-nl, thanks for reaching out (as always). The root certificate looks duplicative (we already have a `-days` value of 730), but others are on the `*-key.pem` lines look like they might indeed be necessary. To your knowledge, the keys also need the longer expiration date?

ict-one-nl:
Hmm I need to test that, expired certs do however bring the cluster down, nodes won't communicate.

Will try to this this week.

aetter:
Thanks a bunch. I'll try to take a look next week, as well. Some of the reading I've done indicates that keys don't expire, but maybe it's just a particular key format, or keys in a different context, etc. Testing will reveal all, I'm sure.

ict-one-nl:
Difficult one to test, but after googling I found everyone using the -days option only on the request and not on the private keys. Think we should be alright with the proposed changes.

aetter:
Hi @ict-one-nl I'm still running some tests here, which take 24 hours apiece while I wait for certs to expire. For keys, agreed, that was my finding, as well. OpenSSL doesn't even accept a `-days` option for them.

Between the CSR and the certificate itself, though, it doesn't make sense to me that they both accept a `-days` argument. It seems like we should only be specifying it in one place. For example, your PR would turn the admin certificate generation into:

```
openssl req -new  -days 3650 -key admin-key.pem -subj "/C=CA/ST=ONTARIO/L=TORONTO/O=ORG/OU=UNIT/CN=ADMIN" -out admin.csr
openssl x509 -req -in admin.csr -CA root-ca.pem -CAkey root-ca-key.pem -CAcreateserial -sha256 -out admin.pem -days 730
```

If we specify `-days` in both, does the 3650 value get overridden by the 730? If not, can we safely remove the 730? That's what I'm looking into.

aetter:
OK I think I can say with confidence that the `-days` argument must go on the final certificate command and not the CSR command. If we use the admin certificate as an example, consider the following:

```
openssl req -new -key admin-key.pem -subj "/C=CA/ST=ONTARIO/L=TORONTO/O=ORG/OU=UNIT/CN=ADMIN" -out admin.csr -days 3650
openssl x509 -req -in admin.csr -CA root-ca.pem -CAkey root-ca-key.pem -CAcreateserial -sha256 -out admin.pem
```

Then run `openssl x509 -enddate -noout -in admin.pem`, and the expiration date is 30 days from today, not ten years. To get the expected ten years, run:

```
openssl req -new -key admin-key.pem -subj "/C=CA/ST=ONTARIO/L=TORONTO/O=ORG/OU=UNIT/CN=ADMIN" -out admin.csr
openssl x509 -req -in admin.csr -CA root-ca.pem -CAkey root-ca-key.pem -CAcreateserial -sha256 -out admin.pem -days 3650
```

So I'm open to changing the expiration date on the script certificates to something other than two years, but the `-days` argument appears to be in the right place. I'm going to close out this PR, but happy to keep discussing in this thread. 👍

ict-one-nl:
Okay cool, I'll update my scripts and see if everything keeps working. Thanks for the effort!

PR #172, opened 2021-09-10
Added security to ISM

keithhc2:
Added ISM to security
 
### Check List


aetter (approved):

aetter:
LGTM

_im-plugin/ism/security.md
```diff
@@ -0,0 +1,40 @@
...
+
+All index state management data are protected as system indices, and only a super admin or an admin with a Transport Layer Security (TLS) certificate can access system indices. For more information, see [System indices]({{site.url}}{{site.baseurl}}/security-plugin/configuration/system-indices).
+
+## Basic permissions
+
+The security plugin comes with two built-in roles that cover most ISM use cases: `ism_read_access` and `ism_full_access`. For descriptions of each, see [Predefined roles]({{site.url}}{{site.baseurl}}/security-plugin/access-control/users-roles#predefined-roles).
+
+With security enabled, users not only need the correct index management permissions, but they also need permissions to relevant indices to execute [REST API]({{site.url}}{{site.baseurl}}/im-plugin/ism/api) calls. For example, if a user wants to use the REST API to attach a policy to an index named `system-logs`, they would need the necessary REST API permissions as well as access to `system-logs`.
```
thalurur, line 19: This is true for all ISM apis except
* Create policy
* Get policy
* Delete policy

Wanted to call out in case you wanted to be more explicit here. Simply put any ISM API that interacts with managed index will require additional permission that user is allowed to manage the index i.e have the permission - `indices:admin/opensearch/ism/managedindex` on index

thalurur (commented):
Should we also callout about the policy action permission requirement?

e.g. after security change if user Jane creates a policy that does say rollover and eventually delete, attaches this policy to some index. The policy will successfully associate as long as Jane as the manage index permission and add policy API permission. 

But for the policy to successfully execute Jane need to have the permission to the do these actions on the index, if not the policy will fail saying the user who created the policy in this case Jane doesn't have permission to do these actions on the index

_im-plugin/security.md
```diff
@@ -0,0 +1,41 @@
...
+
+All index management data are protected as system indices, and only a super admin or an admin with a Transport Layer Security (TLS) certificate can access system indices. For more information, see [System indices]({{site.url}}{{site.baseurl}}/security-plugin/configuration/system-indices).
+
+## Basic permissions
+
+The security plugin comes with one role that offers full access to index management: `index_management_full_access`. For a description of the role's permissions, see [Predefined roles]({{site.url}}{{site.baseurl}}/security-plugin/access-control/users-roles#predefined-roles).
+
+With security enabled, users not only need the correct index management permissions, but they also need permissions to execute actions to involved indices. For example, if a user wants to use the REST API to attach a policy that executes a transform job to an index named `system-logs`, they would need the permissions to attach a policy and execute a transform job, as well as access to `system-logs`.
```
thalurur, line 18: can we say rollup instead of transform we don't have capaiblity to run transform job as part of policy yet?

keithhc2: Changed

thalurur (approved):

Issue #160, opened 2021-08-27
[BUG] Incorrect documentation


abbashus:
The documentation at [security-plugin/access-control/api/configuration](https://opensearch.org/docs/security-plugin/access-control/api/#configuration) and [security-plugin/access-control/api/certificates](https://opensearch.org/docs/security-plugin/access-control/api/#certificates) is same. 

The content under **Certificates** should either be removed or replace with correct documenation.

Also the [Patch configuration REST API](security-plugin/access-control/api/#certificates) does not work. See opensearch-project/security#1428

aetter:
I'm getting extreme deju vu here, @keithhc2. Did we not have/fix something similar...?

keithhc2:
@aetter I thought so, too, but turns out #144 was just about CAs. Either way, I'll comb through the security plugin and see if anything else is amiss.

keithhc2:
Certificates section is now fixed, and patch configuration is updated with a mandatory setting before a user can execute that operation: https://opensearch.org/docs/security-plugin/access-control/api/#patch-configuration

PR #183, opened 2021-09-24
Add descriptions to config yaml files and added some security APIs

### LINKED ISSUE #143: Add documentation for nodes_dn.yml
Currently the `node_dn.yml` is only mentioned in the [YAML files](https://opensearch.org/docs/security-plugin/configuration/yaml/#nodes_dnyml).

The lag of documentation makes it hard to make use of this option.

keithhc2:
Added descriptions to `tenants.yml` and `nodes_dn.yml`, and added `nodes_dn` security APIs.
 
### Issues Resolved
#143 and #160 
 
### Check List

### DIFF
--- a/_security-plugin/access-control/api.md
+++ b/_security-plugin/access-control/api.md
@@ -1288,7 +1288,7 @@ DELETE _plugins/_security/api/nodesdn/<cluster-name>
 Introduced 1.0
 {: .label .label-purple }
 
-Retrieves the cluster's certificates.
+Retrieves the cluster's security certificates.
 
 #### Request
 
--- a/_security-plugin/access-control/api.md
+++ b/_security-plugin/access-control/api.md
@@ -1159,6 +1159,12 @@ Introduced 1.0
 
 Updates the existing configuration using the REST API. This operation can easily break your existing configuration, so we recommend using `securityadmin.sh` instead, which is far safer. See [Access control for the API](#access-control-for-the-api) for how to enable this operation.
 
+Before you can execute the operation, you must first add the following line to `opensearch.yml`:
+
+```yml
+plugins.security.unsupported.restapi.allow_securityconfig_modification: true
+```
+
 #### Request
 
 ```json
--- a/_security-plugin/access-control/api.md
+++ b/_security-plugin/access-control/api.md
@@ -1183,9 +1183,9 @@ PATCH _plugins/_security/api/securityconfig
 
 ## Distinguished names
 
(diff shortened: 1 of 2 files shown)


_security-plugin/access-control/api.md
```diff
@@ -1179,110 +1179,147 @@ PATCH _plugins/_security/api/securityconfig
...
-
 ---
 
-## Certificates
+## Distinguished names
 
-### Get certificates
-Introduced 1.0
-{: .label .label-purple }
+These REST APIs let a super admin allow list distinguished names to enable communication between clusters and/or nodes.
```
aetter, line 1186: So if we want to use the noun form, absolutely, "allow list" seems fine. But you've verbed it here, which makes me think it either needs to be "allowlist" or rephrased to something like "let a super admin list the distinguished names that are allowed to enable communication between clusters and/or nodes."

I get that we already use "allow list" as a verb in a couple places. Check with editorial and make consistent everywhere?

_security-plugin/access-control/api.md
```diff
@@ -1179,110 +1179,147 @@ PATCH _plugins/_security/api/securityconfig
...
+  "cluster1": {
+    "nodes_dn": [
+      "CN=cluster1.example.com"
+    ]
+  }
+}
+```
 
-Creates or updates the existing configuration using the REST API rather than `securityadmin.sh`. This operation can easily break your existing configuration, so we recommend using `securityadmin.sh` instead. See [Access control for the API](#access-control-for-the-api) for how to enable this operation.
+To get the distinguished names from a specific cluster or node, include its name in the request's URL.
```
aetter, line 1217: I think "request path" is clearer than "request's URL," but whatever you think.

_security-plugin/access-control/api.md
```diff
@@ -1179,110 +1179,147 @@ PATCH _plugins/_security/api/securityconfig
...
+{
+  "status": "CREATED",
+  "message": "'cluster3' created."
+}
+```
+
+
+### Delete distinguished names
+
+Deletes the cluster's allow listed distinguished names.
```
aetter, line 1265: Looks like this deletes the entire list? So maybe "Deletes all allow listed distinguished names."?

_security-plugin/access-control/api.md
keithhc2: Turns out we just _don't_ use "allow list" as a verb, so I've changed all "allow list" references to only be nouns.

aetter (approved):

PR #267, opened 2021-11-15
Anomaly result mapping

ashwinkumar12345:
### Description
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List

### DIFF
--- a/_monitoring-plugins/ad/result-mapping.md
+++ b/_monitoring-plugins/ad/result-mapping.md
@@ -1,0 +1,319 @@
+---
+layout: default
+title: Anomaly result mapping
+parent: Anomaly detection
+nav_order: 6
+---
+
+# Anomaly result mapping
+
+If you enabled custom result index, the anomaly detection plugin stores the results in your own index.
+
+If the anomaly detector doesn’t detect an anomaly, the result has the following format:
+
+```json
+{
+  "detector_id": "kzcZ43wBgEQAbjDnhzGF",
+  "schema_version": 5,
+  "data_start_time": 1635898161367,
+  "data_end_time": 1635898221367,
+  "feature_data": [
+    {
+      "feature_id": "processing_bytes_max",
+      "feature_name": "processing bytes max",
+      "data": 2322
+    },
+    {
+      "feature_id": "processing_bytes_avg",
+      "feature_name": "processing bytes avg",
+      "data": 1718.6666666666667
+    },
+    {
+      "feature_id": "processing_bytes_min",
+      "feature_name": "processing bytes min",
+      "data": 1375
+    },
+    {
+      "feature_id": "processing_bytes_sum",
+      "feature_name": "processing bytes sum",
+      "data": 5156
+    },
+    {
+      "feature_id": "processing_time_max",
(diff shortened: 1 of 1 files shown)


keithhc2 (approved):

_monitoring-plugins/ad/result-mapping.md
```diff
@@ -0,0 +1,319 @@
...
+
+If a detector detects an anomaly late, the result has the following additional fields:
+
+Field | Description
+:--- | :---
+`past_values` | The actual input that triggered an anomaly. If `past_values` is null, the attributions or expected values are from the current input. If `past_values` is not null, the attributions or expected values are from a past input (for example, the previous two steps of the data [1,2,3]).
+`approx_anomaly_start_time` | The approximate time of the actual input that triggers an anomaly. This field helps you understand when a detector flags an anomaly. If the data is not continuous, the actual time that the detector detects the anomaly can be earlier. Single-stream detectors query previous anomaly results of a few data points while high-cardinality detectors don't do this because querying previous results is an expensive operation when performed on a lot of entities. So, the accuracy of this field is less for high-cardinality detectors as compared to single-stream detectors.
```
kaituo, line 197: There is no difference between single stream and hcad.  Both of them does not query because of the expensive queries.

lizsnyder (approved):

_monitoring-plugins/ad/result-mapping.md
ashwinkumar12345: Updated. Please review.

_monitoring-plugins/ad/result-mapping.md
```diff
@@ -194,7 +194,7 @@ If a detector detects an anomaly late, the result has the following additional f
...
-`approx_anomaly_start_time` | The approximate time of the actual input that triggers an anomaly. This field helps you understand when a detector flags an anomaly. If the data is not continuous, the actual time that the detector detects the anomaly can be earlier. Single-stream detectors query previous anomaly results of a few data points while high-cardinality detectors don't do this because querying previous results is an expensive operation when performed on a lot of entities. So, the accuracy of this field is less for high-cardinality detectors as compared to single-stream detectors.
+`approx_anomaly_start_time` | The approximate time of the actual input that triggers an anomaly. This field helps you understand when a detector flags an anomaly. If the data is not continuous, the actual time that the detector detects the anomaly can be earlier. Both single-stream and high-cardinality detectors don't query previous anomaly results because these queries are expensive operations, especially when a detector has a lot of entities. This is why the accuracy of this field is low.
```
kaituo, line 197: > "Both single-stream and high-cardinality detectors don't query previous anomaly results because these queries are expensive operations, especially when a detector has a lot of entities. This is why the accuracy of this field is low."
single-stream detector has only 1 entity, only HCAD detectors have a lot of entities. So it is strange to read "Both single-stream .. especially when a detector has a lot of entities." Also, the accuracy is not necessarily low if the data is continuous. 

I wanted to explain why I call the field approximate here: queries are not for free. The cost is especially high for HCAD detectors since they have a lot of entities and there can be a lot of anomalies.

PR #301, opened 2021-11-29
Custom search for 1.0

AMoo-Miki:
Signed-off-by: Miki <mehranb@amazon.com>

### Description
Adds an in-house custom search solution using OS

 
### Check List

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -95,8 +95,15 @@ just_the_docs:
 
 
 # Enable or disable the site search
-# Supports true (default) or false
-search_enabled: true
+# By default, just-the-docs enables its JSON file-based search. We also have an OpenSearch-driven search functionality.
+# To disable any search from appearing, both `search_enabled` and `use_custom_search` need to be false.
+# To use the OpenSearch-driven search, `search_enabled` has to be false and `use_custom_search` needs to be true.
+# If `search_enabled` is true, irrespective of the value of `use_custom_search`, the JSON file-based search appears.
+#
+# `search_enabled` defaults to true
+# `use_custom_search` defaults to false
+search_enabled: false
+use_custom_search: true
 
 search:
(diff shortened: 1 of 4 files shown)


_config.yml
```diff
@@ -94,9 +94,9 @@ just_the_docs:
       name: External links
 
 
-# Enable or disable the site search
+# Enable or disable the off-the-shelf site search
 # Supports true (default) or false
-search_enabled: true
+search_enabled: false
```
ashwin-pc, line 99: Can we rename this flag to be more specific? It confusing to parse the logic later in `default.html` what this flag does because we have a search bar irrespective of the flag being true or false

AMoo-Miki: This is defined by the underlying application we use for the documentation: https://github.com/pmarsceill/just-the-docs. Do you feel more explanation is warranted?

_layouts/default.html
```diff
@@ -213,5 +220,8 @@ <h2 class="text-delta">Table of contents</h2>
     </script>
   {% endif %}
   <script src="{{ '/assets/js/header-nav.js' | relative_url }}"></script>
+  {% if site.search_enabled != false and site.use_custom_search != false %}
```
abbashus, line 223: I was thinking if checking for `true` made more sense here?
Something like 

```
 {% if site.search_enabled and site.use_custom_search %}
```

AMoo-Miki: JTD defaults `search_enabled` to true and hence we are forced to use `== false` or `!= false` as the value of `false` is the only thing that turns it off.

abbashus (approved):
Thanks for the changes!

ashwin-pc (approved):
LGTM!

ashwin-pc (approved):

ashwinkumar12345 (approved):

Issue #389, opened 2022-01-31
[Documentation/UX Update] index_pattern/s clarification


dtaivpp:
Hey All,

Putting this here as I am not sure if this should be classified as a bug as the documentation on it is not clear. For OpenSearch-Dashboards the [documentation](https://opensearch.org/docs/latest/opensearch/index-templates/) states, "to any new index whose name matches the regular expression logs-2020-01-*" implying that regular expressions can be used. This seems to be bad wording as the field only seems to support asterisks as wild cards. (also in Open Distro's [documentation](https://opendistro.github.io/for-elasticsearch-docs/docs/elasticsearch/index-templates/#index-template))

This is also confirmed on the page for creating index patterns which states `Use an asterisk (*) to match multiple indices. Spaces and the characters \, /, ?, ", <, >, | are not allowed.`

The documentation should reflect the hint above for this. Also, if this pattern is the same for ISM then it should be reflected in their [documentation](https://opensearch.org/docs/latest/im-plugin/ism/index/).

kavilla:
Thanks for opening! Yeah I agree the docs implies regex accepted. Moved to doc repo.

keithhc2:
Updated with #394

PR #394, opened 2022-02-04
Adjusted index pattern language

keithhc2:
#389 
 
### Check List


Naarcha-AWS (approved):

lizsnyder (approved):

lizsnyder:
Should we agree to use "indexes" rather than "indices" in all our docs going forward?

Naarcha-AWS:
> Should we agree to use "indexes" rather than "indices" in all our docs going forward?

I think so. "Indexes" reads as more friendly than "Indices", which has a formal tone. Also most developer-focused docs, including Azure, uses "Indexes".

keithhc2:
Yeah, not to mention editorial says we should use "indexes", so we should be consistent with docs over on the AWS side.

Issue #395, opened 2022-02-05
[BUG] k-NN endpoint is incorrect


vishalsrao:
The following command on https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/ is incorrect:
```
POST /_plugins/_knn/models/_train/my-model
```

I think it should be:
```
POST /_plugins/_knn/models/my-model/_train
```

Additionally, the index train-index should probably contain documents with higher embedding-dimensions to be compatible with pq parameter m=8.

It would be good to have a link to an example which can be used to load more embeddings. PQ requires at least 256 samples to be compatible with code_size 8.

Naarcha-AWS:
I suspect that this might have been fixed due to recent updates in the kNN documentation. Will investigate and close if so.

hdhalter:
@cwillum - Looks like the first issue is fixed (_train). Can you please see if the second request has been addressed? Thanks.

hdhalter:
@martin-gaievski - following up on this inquiry. @kolchfa-aws - for visibility, since I know you've reworked this section. Thanks!

martin-gaievski:
PR author is correct, url must be `/_plugins/_knn/models/my-model/_train`, that is a bug and needs to be corrected

hdhalter:
> PR author is correct, url must be `/_plugins/_knn/models/my-model/_train`, that is a bug and needs to be corrected

This has been updated in the documentation already: POST /_plugins/_knn/models/my-model/_train. 

The question is on the second part of their suggestion:  
"Additionally, the index train-index should probably contain documents with higher embedding-dimensions to be compatible with pq parameter m=8.

It would be good to have a link to an example which can be used to load more embeddings. PQ requires at least 256 samples to be compatible with code_size 8."

Any input you can provide here is appreciated. Thanks!

hdhalter:
The second item is addressed here: https://opensearch.org/docs/latest/search-plugins/knn/knn-index#supported-faiss-encoders

Issue #393, opened 2022-02-04
[Documentation] ISM Actions


dtaivpp:
Hey All, 

It's come to our attention that the [Index Policy Actions documentation](https://opensearch.org/docs/latest/im-plugin/ism/policies/#actions) needs to be clarified. 

The ISM actions will execute synchronously even if there is a timeout specified. The timeout acts as a circuit breaker rather than a buffer period between actions as implied by the current documentation. 

[https://github.com/opensearch-project/index-management/blob/6141c3c6fec9fe4c0c7bf7[…]arch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt](https://github.com/opensearch-project/index-management/blob/6141c3c6fec9fe4c0c7bf733877660e78ed2b1b1/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt#L282)

Documentation should be upgraded to reflect this.

keithhc2:
Fixed with #412

PR #408, opened 2022-02-11
General bug fixes

### LINKED ISSUE #404: [DOCUMENTATION] the example in documentation has invalid package paths
the documentation on https://opensearch.org/docs/latest/clients/java/ has imports with the package  `org.opensearch.clients` which does not exist. Seems the correct package is  `org.opensearch.client`.

keithhc2:
#404 #395 #393 

### Check List


dblock:
Thanks. Is the code given as example in this post executing anywhere with CI? It could be in the client's unit tests, a standalone sample in this repo, etc. Either way I think we should take the habit of doing that so when the code changes we ensure 100% of the time that it works.

keithhc2:
No, our current code examples don't execute with CI. I do agree that it would be a good idea to have some mechanic in place that automatically updates our code samples whenever a version, of say, the clients release, but we'll need support from engineering for that, especially since they've taken a bigger lead on producing new code example pages for newly released clients.

PR #483, opened 2022-03-31
Add Basic Documentation for Terms Lookup Query

### LINKED ISSUE #430: [FEATURE] documentation for terms lookup queries
**Is your feature request related to a problem?**
opensearch supports terms lookup queries, but they're currently undocumented.

**What solution would you like?**
extend the [terms query documentation](https://opensearch.org/docs/latest/opensearch/query-dsl/term/#terms) to include terms lookup queries.

**What alternatives have you considered?**
n/a

**Do you have any additional context?**
see the original [documentation from ES](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/query-dsl-terms-query.html#query-dsl-terms-lookup).

rursprung:
### Description
this adds very rudimentary documentation for TLQ so that it at least
doesn't go unmentioned. however, there are many more details which
should be mentioned. this should then be done with #430.

more information can be found in the [Elasticsearch TLQ
Documentation][ESTLQ]. i'm not 100% sure whether this was under the
Apache 2.0 license in 7.10 (the repo states that everything is Apache
2.0 unless explicitly stated otherwise in the file and the documentation
files don't seem to include any other license?) and thus didn't simply
take over the documentation from there (which would of course be the
easiest and most complete way of doing it).

also i'm not sure (and didn't find this information anywhere) whether
the test data actually exists and the queries can be run against the
test data? in that case the provided example either needs to be adapted
to the existing test data or the test data needs to be extended
accordingly (i doubt that an index `play-assignments` already exists?).
due to this the example has also not been tested.

Signed-off-by: Ralph Ursprung <Ralph.Ursprung@avaloq.com>

[ESTLQ]: https://www.elastic.co/guide/en/elasticsearch/reference/7.10/query-dsl-terms-query.html#query-dsl-terms-lookup
 
### Issues Resolved
partially resolves #430 (but more will be required to fully resolve it, but i'll leave that to someone else with better knowledge about TLQ as the ES docs show that it has some details to be aware of)
 
### Check List

### DIFF
--- a/_opensearch/query-dsl/term.md
+++ b/_opensearch/query-dsl/term.md
@@ -271,6 +271,33 @@ GET shakespeare/_search
 
 You get back documents that match any of the terms.
 
+### Terms Lookup
+
+You can use a `terms` query with a `lookup` to match values based on a field in a document in another index.
+
+Parameter | Behavior
+:--- | :---
+`index` | The index from which the document is read.
+`id` | The id of the documented.
+`path` | Path to the field from which the values are used in the terms query.
+
+E.g. to get all lines from the shakespeare play for a role (or roles) specified in the index `play-assignments` for the entry `42`:
+
+```json
+GET shakespeare/_search
+{
+  "query": {
+    "terms": {
+      "speaker": {
+        "index": "play-assignments",
+        "id": "42",
+        "path": "role"
+      }
+    }
+  }
+}
+```
+
 ## IDs
 
 Use the `ids` query to search for one or more document ID values.


rursprung:
> Hey @rursprung. We'll have to modify this more to make it OpenSearch specific. As a general rule of thumb, we can't reuse any documentation from a third-party, even under an Apache 2.0 license.

just for clarification: i wrote 100% of the content of this PR w/o taking it over from ES (i actually refrained from re-reading their docs before writing this PR to avoid accidentally taking anything over and only controlled it afterwards) which is also why it's incomplete.

> Would you be okay if I took over this PR from here?

absolutely, go ahead! thanks a lot!

alicejw-aws:
Hi @rursprung ,
I also have some changes to the same section 
( Query DSL) that will compliment your changes for this PR. I've made some minor suggestions in your PR that align with the other changes. 
Please take a look at this PR: https://github.com/opensearch-project/documentation-website/pull/561
I hope it makes sense to view both PR changes together. Thanks

_opensearch/query-dsl/term.md
```diff
@@ -271,6 +271,33 @@ GET shakespeare/_search
...
 You get back documents that match any of the terms.
 
+### Terms Lookup
+
+You can use a `terms` query with a `lookup` to match values based on a field in a document in another index.
+
+Parameter | Behavior
+:--- | :---
+`index` | The index from which the document is read.
+`id` | The id of the documented.
```
keithhc2, line 281: Typo here. I'm guessing it should be "document" instead of "documented"?

keithhc2 (approved):

lizsnyder (approved):

_opensearch/query-dsl/term.md
```diff
@@ -271,6 +271,33 @@ GET shakespeare/_search
...
+`path` | Path to the field from which the values are used in the terms query.
+
+E.g. to get all lines from the shakespeare play for a role (or roles) specified in the index `play-assignments` for the entry `42`:
```
lizsnyder, line 284: Tiny nit, maybe capitalize Shakespeare

alicejw-aws (commented):
Hi @rursprung ,
I also have some changes to the same section
( Query DSL) that will compliment your changes for this PR. I've made some minor suggestions in your PR that align with the other changes.
Please take a look at this PR: https://github.com/opensearch-project/documentation-website/pull/561
I hope it makes sense to view both PR changes together. Thanks

rursprung:
@alicejw-aws: since you're editing the exact same files i guess this would lead to merge conflicts for the 2nd PR after merging the first (regardless of which one goes in first). it might make sense for you to just take over the changes i proposed here directly into your PR to avoid having to rebase / merge afterwards?

_opensearch/query-dsl/term.md
```diff
@@ -271,6 +271,33 @@ GET shakespeare/_search
 
 You get back documents that match any of the terms.
 
+### Terms Lookup
```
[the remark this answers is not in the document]
rursprung: IMHO this is wrong. there's a difference between `term` queries and `terms` queries and only `terms` supports `lookup`, so i would strongly advise against this change in nomenclature!

alicejw-aws:
Hi @rursprung, thanks for looking out for merge conflicts, i do appreciate that. However, the changes I suggested here for you are separate (yet complimentary and need to be entered either in this PR or in a separate one). 

If you prefer to have a separate PR for my changes to this file, I can do that. (i was trying to save time and effort by doing it with this PR).

alicejw-aws:
Ok. The plan is to merge this PR without my proposed changes, then add my changes to the separate [561 PR](https://github.com/opensearch-project/documentation-website/pull/561)

no worries. this should work out with the same outcome: consistent complimentary information that contains the uniform terms throughout the doc set (e.g. Term Lookup Query TQL).

rursprung:
ah, i think we misunderstood each other: my proposal was to just abandon this PR and take over everything you want from this directly in #561 - not to merge this one here w/o your suggestion (otherwise i'd have applied them)

alicejw-aws:
oh ok thanks. no worries. i acted on your previous response, so it's merged now. my PR is in-review for the changes i had suggested previously. hopefully i can get those merged today and it will be ok.

Issue #591, opened 2022-05-20
[BUG] Nodes APIs introduction fixes


lukas-vlcek:
### What is the bug?

After [#584] has been merge couple of issues appeared:

**First**, dead links were detected:
```
LinkChecker: [Warning] Found 6 dead links:
http://localhost:4000/docs/latest/opensearch/nodes-apis/index/, linked to in ./_opensearch/rest-api/cluster-stats.md
../index/#node-filters, linked to in ./_opensearch/rest-api/nodes-apis/nodes-hot-threads.md, ./_opensearch/rest-api/nodes-apis/nodes-info.md
../index/#timeout, linked to in ./_opensearch/rest-api/nodes-apis/nodes-hot-threads.md, ./_opensearch/rest-api/nodes-apis/nodes-info.md
../../../configuration/#configuration-file, linked to in ./_opensearch/rest-api/nodes-apis/nodes-info.md
../../../configuration/#update-cluster-settings-using-the-api, linked to in ./_opensearch/rest-api/nodes-apis/nodes-info.md
../../../aggregations, linked to in ./_opensearch/rest-api/nodes-apis/nodes-info.md
                    ...done in 7.032533 seconds.
```

**Second**, all child nodes of newly introduced "Nodes APIs" parent were not included in the content. I.e. child pages "Nodes info" and "Nodes hot threads" are not available via navigation menu.

### How can one reproduce the bug?

Checkout the merge commit:
```
% git log -n 1
commit 08d8125f43d1083b64d3f8d4a7cf5f220a3640e3 (HEAD -> main, origin/main, origin/HEAD)
Merge: 099ba00 e437eaa
Author: Naarcha-AWS <97990722+Naarcha-AWS@users.noreply.github.com>
Date:   Thu May 19 10:26:02 2022 -0500

    Merge pull request #584 from opensearch-project/lukas-vlcek-add_node_APIs
    
    Add Node APIs [Edits to Lukas Vlcek] 
```
and generate the site:
```
% sh ./build.sh
```

### What is the expected behaviour?

No dead links and all pages rendered correctly. Especially pages under Nodes APIs parent.

PR #592, opened 2022-05-20
Fixing Nodes APIs issues

lukas-vlcek:
### Fixing Nodes APIs issues

1. Fixing all reported dead links.
2. Changing `title` value in index.md back to plural form, i.e. "Nodes APIs".

The second change fixed missing child pages. I am not saying I insist on use of plural form but if we want to change it to a singular form then we also have to change `parent` fields in all child pages to correctly refer to valid parent, otherwise all child pages and content is missing.

As for the debate of use APIs vs API (plural vs singular) I just want to point out that right now there are many other parent pages using plural form. If we want to change that then I suggest doing it in separate PR and include other relevant parent pages too.

### Issues Resolved

Closes: #591

Signed-off-by: Lukáš Vlček <lukas.vlcek@aiven.io>

### DIFF
--- a/_opensearch/rest-api/nodes-apis/index.md
+++ b/_opensearch/rest-api/nodes-apis/index.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: Nodes API
+title: Nodes APIs
 parent: REST API reference
 has_children: true
 nav_order: 5
--- a/_opensearch/rest-api/cluster-stats.md
+++ b/_opensearch/rest-api/cluster-stats.md
@@ -31,7 +31,7 @@ All cluster stats parameters are optional.
 
 Parameter | Type | Description
 :--- | :--- | :---
-&lt;node-filters&gt; | List | A comma-separated list of [node filters](../nodes-apis/index/#node-filters) that OpenSearch uses to filter results.
+&lt;node-filters&gt; | List | A comma-separated list of [node filters]({{site.url}}{{site.baseurl}}/opensearch/rest-api/nodes-apis/index/#node-filters) that OpenSearch uses to filter results.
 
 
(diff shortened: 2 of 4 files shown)


lukas-vlcek:
FYI @Naarcha-AWS @alicejw-aws @lizsnyder @natebower
Thank you very much 🙏

_opensearch/rest-api/nodes-apis/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: Nodes API
+title: Nodes APIs
```
Naarcha-AWS, line 3: We should use either "Node APIs" or "Nodes API" for grammatical purposes. Since this does encompass multiple endpoints, let's use "Node APIs".

Naarcha-AWS (commented):
Quick change, then we should be good.

_opensearch/rest-api/nodes-apis/index.md
lukas-vlcek: I am fine with that, this also means that we need to update `parent` field in all child pages.

Naarcha-AWS: Fair point on the REST API. I'm fine with it matching and not worrying about the language aspect if you are. :D

Naarcha-AWS (approved):

lizsnyder (approved):

_opensearch/rest-api/nodes-apis/index.md
lukas-vlcek: Yes, I think so.
The redirect should be IMO discussed as a general topic for the documentation.

PR #912, opened 2022-08-18
Fix setting name for LDAP custom attribute filter

peternied:
### Description
Fix setting name for LDAP custom attribute filter to the setting name used in the security plugin

### Issues Resolved
* Related to https://github.com/opensearch-project/security/issues/2032

### DIFF
--- a/_security-plugin/configuration/ldap.md
+++ b/_security-plugin/configuration/ldap.md
@@ -433,7 +433,7 @@ By default, the security plugin reads all LDAP user attributes and makes them av
 
 Name | Description
 :--- | :---
-`custom_attr_allowlist`  | String array. Specifies the LDAP attributes that should be made available for variable substitution.
+`custom_attr_whitelist`  | String array. Specifies the LDAP attributes that should be made available for variable substitution.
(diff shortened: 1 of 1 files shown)


cwillum (commented):
@peternied It looks like this will be a temporary change to documentation until the key in the code is updated to adhere to the OS [style sheet](https://github.com/opensearch-project/documentation-website/blob/main/STYLE_GUIDE.md) and inclusive content. Should there be another GitHub issue (or does one already exist) aimed at updating the code so we can keep an eye on this and revert the documentation back to `custom_attr_allowlist` when it's available?
Adding @natebower for awareness.

peternied:
@cwillum Good question, https://github.com/opensearch-project/security/issues/2032 will likely be converted into 'name the configuration correctly', but the legacy name will need to be supported and in the documentation until the 3.0.0 release.  I'm not sure if we have a good mechanism for tracking this in the docs site, but we might have more issues like this that need to be revisited

cwillum (approved):
@peternied Sounds good. Will keep in mind that this might need attention again in 3.0.0.

Naarcha-AWS (approved):

PR #919, opened 2022-08-18
Missing documentation for the plugins.security.allow_default_init_securityindex property

### LINKED ISSUE #676: [FEATURE] missing documentation on configuration: plugins.security.allow_default_init_securityindex
this isn't a new feature, it has been around for a long time.

the `plugins.security.allow_default_init_securityindex` config option must be documented. by default it's `false`. when set to `true` the configuration for the security plugin is loaded from the config folder on initial startup if the node can't find a security configuration index.

n/a

* for details please check with the team behind [opensearch/security](https://github.com/opensearch-project/security)
* the config option is defined here:
  https://github.com/opensearch-project/security/blob/e40d9358077440494322685b970259e628b3b0eb/src/main/java/org/opensearch/security/support/ConfigConstants.java#L214
* Search Guard has a very similar feature which is partially documented here: https://docs.search-guard.com/latest/demo-installer-generated-artefacts#auto-initialization

cwillum:

Fixes #676 

### Description
Adds description to the "YAML files" section of security plugin documentation for the `plugins.security.allow_default_init_securityindex` property.

### DIFF
--- a/_security-plugin/configuration/yaml.md
+++ b/_security-plugin/configuration/yaml.md
@@ -124,7 +124,7 @@ plugins.security.restapi.password_validation_regex: '(?=.*[A-Z])(?=.*[^a-zA-Z\d]
 plugins.security.restapi.password_validation_error_message: "Password must be minimum 8 characters long and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character."
 ```
 
-The opensearch.yml file also contains the `plugins.security.allow_default_init_securityindex` property. When set to `true`, the security plugin uses default security settings if an attempt to create the security index fails when OpenSearch launches. Default security settings are stored in YAML files contained in the config folder. By default, this setting is `false`.
+The opensearch.yml file also contains the `plugins.security.allow_default_init_securityindex` property. When set to `true`, the security plugin uses default security settings if an attempt to create the security index fails when OpenSearch launches. Default security settings are stored in YAML files contained in the `opensearch-project/security/config` directory. By default, this setting is `false`.
 
 ```yml
 plugins.security.allow_default_init_securityindex: true
--- a/_security-plugin/configuration/yaml.md
+++ b/_security-plugin/configuration/yaml.md
@@ -124,7 +124,7 @@ plugins.security.restapi.password_validation_regex: '(?=.*[A-Z])(?=.*[^a-zA-Z\d]
 plugins.security.restapi.password_validation_error_message: "Password must be minimum 8 characters long and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character."
 ```
 
-You can set the `plugins.security.allow_default_init_securityindex` property to `true` to have the security plugin automatically initialize with default settings in the config folder if the security index fails to initialize when OpenSearch launches. By default, the setting is `false`.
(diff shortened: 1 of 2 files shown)


cwillum:
@cliu123 Could you share your thoughts on a couple of things? I'm wondering if we should be a little more specific about what the "default settings" in the config folder are about. Are these contained in specific files? What settings do they include? Or is this enough explanation for the property?

_security-plugin/configuration/yaml.md
```diff
@@ -124,6 +124,12 @@ plugins.security.restapi.password_validation_regex: '(?=.*[A-Z])(?=.*[^a-zA-Z\d]
 plugins.security.restapi.password_validation_error_message: "Password must be minimum 8 characters long and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character."
 ```
 
+You can set the `plugins.security.allow_default_init_securityindex` property to `true` to have the security plugin automatically initialize with default settings in the config folder if the security index fails to initialize when OpenSearch launches. By default, the setting is `false`.
```
cliu123, line 127: When this setting is `true`, security plugin will attempt to create security index and default configs if they are absent.

cliu123, line 127: Just want to make sure the comment here conveys the right message. When this setting is true, security plugin will attempt to create security index and default configs if they are absent.

cliu123:
> @cliu123 Could you share your thoughts on a couple of things? I'm wondering if we should be a little more specific about what the "default settings" in the config folder are about. Are these contained in specific files? What settings do they include? Or is this enough explanation for the property?

@cwillum Sorry for responding to this late! The default settings are here: https://github.com/opensearch-project/security/tree/main/config

cwillum:
Reworded for better clarity.

_security-plugin/configuration/yaml.md
```diff
@@ -124,6 +124,12 @@ plugins.security.restapi.password_validation_regex: '(?=.*[A-Z])(?=.*[^a-zA-Z\d]
 plugins.security.restapi.password_validation_error_message: "Password must be minimum 8 characters long and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character."
 ```
 
+The opensearch.yml file also contains the `plugins.security.allow_default_init_securityindex` property. When set to `true`, the security plugin uses default security settings if an attempt to create the security index fails when OpenSearch launches. Default security settings are stored in YAML files contained in the config folder. By default, this setting is `false`.
```
JeffH-AWS, line 127: I would clarify which "config" folder you mean. If it's `opensearch/config` then it might be good to call it out as a relative path like that. Or if the reader already knows from somewhere earlier in the document that "the config folder" is located there, then maybe add ticks so it's monospaced like `config`.

JeffH-AWS: Also, need to change it to `directory` instead of `folder`.  We don't have folders in Linux.

alicejw-aws (approved):
Looks good

_security-plugin/configuration/yaml.md
cwillum: Incorporated those suggestions. Good improvements.

Naarcha-AWS (approved):

PR #1054, opened 2022-08-31
Adds disclaimer about Maven coords URL formation and details about local zip files

### LINKED ISSUE #1051: [DOC] Add information about Maven coordinate limitations
**What do you want to do?**

From comments in PR #1041:

```
what's maybe missing here is the information that for local files you have to use file:/// as a prefix for the path, i.e. you can't just write /tmp/some-plugin.zip but have to write file:////tmp/some-plugin.zip (or, under windows - once it's supported -: file:///C:\temp\some-plugin.zip)
```

```
it's maybe important to add a note about the restriction that this only goes against maven central (it doesn't actually use maven to find the artefact but it just constructs the URL in the background), so there's (currently) no way to e.g. configure it to go against a proxy maven server instead (as you could do with normal maven builds).

it can well be that in a build environment there's no access to the internet but there is access to internal mirrors. accordingly, one might think that with this method it'd be possible to access it from the maven mirror (because the proper maven setting is defined for the user running the command).
```

JeffH-AWS:
### Description
Added clarification for readers about how Maven coordinates are used so that it's very clear that a host needs connectivity to Maven Central in order to pull down a plugin using this method.

Also added a blurb about installing from a local zip, and mentioned that the only protocols supported are HTTP and HTTPS (no s3 or s3ssl, for example).

### Issues Resolved
Fixes #1051 

### DIFF
--- a/_opensearch/install/plugins.md
+++ b/_opensearch/install/plugins.md
@@ -80,18 +80,43 @@ $ sudo ./opensearch-plugin install analysis-icu
 
 ### Install a plugin from a zip file:
 
+For zip files hosted on a remote server, replace `<zip-file>` with the URL of the hosted file. For zip files in a local directory, replace `<zip-file>` with `file:` followed by the absolute or relative path to the plugin zip file.
+
 #### Usage:
 ```bash
-# zip-file can be a local path to the zip file, or
-# a URL for a zip file hosted on a network.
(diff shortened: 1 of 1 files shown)


Naarcha-AWS (approved):

alicejw-aws (approved):
Looks good!

PR #1173, opened 2022-09-11
Update join.md

xluo-aws:
### Description
Remove mappings object from request body because _mappings are already specified in the request url. However if the index doesn't exist yet, you should remove _mappings from request url and add mappings object inside request body otherwise no index found exception would be thrown

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

Signed-off-by: Xuesong Luo <lxuesong@amazon.com>


xluo-aws:
Didn't commit with DCO information, will submit another PR

PR #1175, opened 2022-09-11
Remove _mappings from create mapping url

xluo-aws:
_mappings and mappings should not be specified at the same time when sending out create mapping request If index doesn't exist, the url should not include _mappings as part of it's path, otherwise it will throw index not exist error If index already exists, the mappings object should not be included in the body, other wise it will throw mapping parse error exception

Signed-off-by: Xuesong Luo <lxuesong@amazon.com>

### Description
_Describe what this change achieves._

### Issues Resolved
https://github.com/opensearch-project/documentation-website/issues/1174


xluo-aws:
Should use my github name instead of real name to commit DCO info. Will create another PR

Issue #1583, opened 2022-10-17
Add documentation for pool pruning period and idle time settings for LDAP


cwillum:
Add documentation that covers configuration of LDAP connection pool pruning period and idle time in the config.yml file. This references issue [#2090](https://github.com/opensearch-project/security/issues/2090) in the security repo.

hdhalter:
Closing, as PR was merged.

PR #1698, opened 2022-10-26
Add connection pooling settings to LDAP documentation

cwillum:

### Description
Upon request to add documentation for new pool pruning period and pool pruning idle time settings to LDAP documentation, it became apparent that all connection pooling settings keys needed to be added to LDAP documentation.

### Issues Resolved

Fixes issue #1583.

Added connection pooling settings to LDAP documentation:
pool.enabled: true
pool.min_size: 0
pool.max_size: 5
pool.idle_time: 2
pool.pruning_period: 1

Also created a new heading for advanced settings (which includes connection pooling settings) so that essential configuration for connection, authentication, and authorization are clearly distinct from these optional settings.

See PR [#2091](https://github.com/opensearch-project/security/pull/2091) in security for original request to add the settings to the code.

### Checklist

### DIFF
--- a/_security-plugin/configuration/ldap.md
+++ b/_security-plugin/configuration/ldap.md
@@ -643,7 +643,7 @@ authz:
 
 ### Connection pooling settings
 
-The security plugin can maintain a pool of connections at the ready, assigning them when needed and returning them to the pool after a connection is closed. This arrangement can lower demands on the resources used to create connections, improve OpenSearch performance, and reduce load on the server. You can use the settings below to control the way the pooling is carried out. 
+OpenSearch can maintain a pool of connections at the ready, assigning them when needed and returning them to the pool after a connection is closed. This arrangement can lower demands on the resources used to create connections, improve OpenSearch performance, and reduce load on the server. You can use the settings below to control the way connection pooling is carried out.
 
 Name | Description
 :--- | :---
--- a/_security-plugin/configuration/ldap.md
+++ b/_security-plugin/configuration/ldap.md
@@ -11,7 +11,6 @@ Active Directory and LDAP can be used for both authentication and authorization 
 
 In most cases, you want to configure both authentication and authorization. You can also use authentication only and map the users retrieved from LDAP directly to security plugin roles.
 
-Active Directory 
 
 ## Docker example
 
--- a/_security-plugin/configuration/ldap.md
+++ b/_security-plugin/configuration/ldap.md
@@ -428,7 +428,7 @@ rolesearch_enabled: false
 
 ## Advanced settings
 
-The advanced settings presented below are optional for an essential LDAP configuration. They can, however, improve efficiency and performance for the LDAP implementation. 
+The advanced settings presented below are optional for an essential LDAP configuration. They can, however, improve efficiency, performance, and security for the LDAP implementation. 
 
 ### Control LDAP user attributes
 
@@ -643,7 +643,7 @@ authz:
 
 ### Connection pooling settings
 
-The directory server can maintain a pool of connections at the ready, assigning them when needed and returning them to the pool after a connection is closed. This arrangement can lower demands on the resources used to create connections, improve OpenSearch performance, and reduce load on the server. You can use the settings below to control the way the pooling is carried out. 
+The security plugin can maintain a pool of connections at the ready, assigning them when needed and returning them to the pool after a connection is closed. This arrangement can lower demands on the resources used to create connections, improve OpenSearch performance, and reduce load on the server. You can use the settings below to control the way the pooling is carried out. 
 
 Name | Description
 :--- | :---
--- a/_security-plugin/configuration/ldap.md
+++ b/_security-plugin/configuration/ldap.md
@@ -429,13 +429,13 @@ rolesearch_enabled: false
 
 ## Advanced settings
 
-The advanced settings presented below are largely optional, but they can improve efficiency and performance for the LDAP-based authentication and authorization configuration. 
+The advanced settings presented below are optional for an essential LDAP configuration. They can, however, improve efficiency and performance for the LDAP implementation. 
 
 ### Control LDAP user attributes
 
 By default, the security plugin reads all LDAP user attributes and makes them available for index name variable substitution and DLS query variable substitution. If your LDAP entries have a lot of attributes, you might want to control which attributes should be made available. The fewer the attributes, the better the performance.
 
-Note that this setting is made in the authentication `authc` section of the config.yml file.
+Note that this setting is made in the `authc` section of the config.yml file.
 
 Name | Description
 :--- | :---
@@ -459,7 +459,6 @@ authc:
       ...
 ```
 
-
 ### Exclude certain users from role lookup
 
 If you are using multiple authentication methods, it can make sense to exclude certain users from the LDAP role lookup.
@@ -477,7 +476,6 @@ skip_users:
   - '/\S*/'
 ```
 
-
 ### Exclude roles from nested role lookups
 
 If the users in your LDAP installation have a large number of roles, and you have the requirement to resolve nested roles as well, you might run into performance issues.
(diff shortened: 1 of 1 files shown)


cwillum:
@opensearch-project/security Could someone on the team have a look at this for technical accuracy? Thanks.

_security-plugin/configuration/ldap.md
```diff
@@ -473,8 +475,7 @@ skip_users:
   - '/\S*/'
 ```
 
-
-### (Advanced) Exclude roles from nested role lookups
+### Exclude roles from nested role lookups
 
 If the users in your LDAP installation have a large number of roles, and you have the requirement to resolve nested roles as well, you might run into performance issues.
```
vagimeli, line 480: Suggested rewrite to tighten wording: If the users in your LDAP installation have several roles and you have requirements to resolve nested roles, you might encounter performance issues.

_security-plugin/configuration/ldap.md
```diff
@@ -540,7 +539,7 @@ authz:
           - '/\S*/'
 ```
 
-### (Advanced) Configuring multiple user and role bases
+### Configuring multiple user and role bases
 
 To configure multiple user bases in the authc and/or authz section, use the following syntax:
```
vagimeli, line 544: Change and/or to "or"

vagimeli (approved):

peternied (approved):

_security-plugin/configuration/ldap.md
```diff
@@ -641,3 +640,33 @@ authz:
...
+
+The directory server can maintain a pool of connections at the ready, assigning them when needed and returning them to the pool after a connection is closed. This arrangement can lower demands on the resources used to create connections, improve OpenSearch performance, and reduce load on the server. You can use the settings below to control the way the pooling is carried out. 
+
+Name | Description
+:--- | :---
+`pool.enabled` | Enables connection pooling. Set to `true` to enable.
+`pool.min_size` | Size of the pool at initialization. Also used as a lower limit when pruning.
+`pool.max_size` | Maximum size the pool can reach.
+`pool.pruning_period` | The interval in minutes at which the pruning implementation is executed. For example: when 5, the implementation is executed every five minutes. By default, the period is 5.
+`pool.idle_time` | The length of time elapsed, in minutes, after a connnection is considered idle. Once elapsed, the connection becomes a candidate for pruning from the pool. By default, idle time is 10.
```
peternied, line 654: These settings are correct for v2.4+ otherwise they have different values `pruning.period` and `pruning.idleTime` in the older versions.

In 2.4+ legacy settings are still read and used and they are marked as deprecated

_security-plugin/configuration/ldap.md
cwillum: @vagimeli Yes, good suggestion. I can incorporate that here. There is, however, another PR for general LDAP updates, and I'll focus on polishing language there. This PR is mainly for pooling settings.

_security-plugin/configuration/ldap.md
cwillum: @peternied I'm not sure I understand what the differences will be between versions -2.4 and versions 2.4+. Will all five connection pooling settings be available to versions before 2.4? If yes, I think you're saying the values will be different for the `pruning.period` and `pruning.idleTime` settings. If that's correct, what will they be?

Ultimately, I'm wondering what details in the documentation added here will lead users astray when they add these settings to config.yml and they're running versions before 2.4? If there are differences, it sounds like I'll need to create a separate PR for earlier versions.

peternied: ### Versions Less than 2.4.0 (e.g. 1.3, 2.1, 2.2, 2.3)
`pruning.period` and `pruning.idleTime` work, but were never documented except for the code

### Versions Equal or greater than 2.4.0
`pool.idle_time` and `pool.pruning_period` work.
`pruning.period` and `pruning.idleTime` work and spit out a deprecation warning in the OpenSearch log

I think you could document all the settings for the version to release in 2.4.0.  Then make another pull request that removes them targeting the older versions.  Someone familiar with this repo might have cleaner instructions or a workflow to follow

_security-plugin/configuration/ldap.md
```diff
@@ -426,12 +426,15 @@ If you don't use or have a role subtree, you can disable the role search complet
 rolesearch_enabled: false
 ```
 
+## Advanced settings
```
cwperks, line 429: 👍

_security-plugin/configuration/ldap.md
```diff
@@ -426,12 +426,15 @@ If you don't use or have a role subtree, you can disable the role search complet
...
 
-### (Advanced) Control LDAP user attributes
+The advanced settings presented below are optional for an essential LDAP configuration. They can, however, improve efficiency and performance for the LDAP implementation.
```
cwperks, line 431: These settings also improve security as well. Perhaps, `efficiency, performance and security`?

_security-plugin/configuration/ldap.md
```diff
@@ -641,3 +640,33 @@ authz:
         rolename: cn
         resolve_nested_roles: true
 ```
+
+### Connection pooling settings
+
+The directory server can maintain a pool of connections at the ready, assigning them when needed and returning them to the pool after a connection is closed. This arrangement can lower demands on the resources used to create connections, improve OpenSearch performance, and reduce load on the server. You can use the settings below to control the way the pooling is carried out.
```
cwperks, line 646: Opensearch is the one maintaining the pool of connections, not the directory server. The security plugin achieves pooling using this library: https://www.ldaptive.org/docs/guide/connections/pooling.html

I'm also only seeing connection pooling in `LDAPAuthenticationBackend2` here: https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/ldap2/LDAPAuthenticationBackend2.java#L71

Diving in deeper now to determine when `LDAPAuthenticationBackend2` is used instead of `LDAPAuthenticationBackend` and will follow-up with another comment when I determine when each is used.

cwperks (commented):
Overall looks good to me. Looking into the different between `LDAPAuthenticationBackend2` and `LDAPAuthenticationBackend` and will follow-up with a more certain answer.

_security-plugin/configuration/ldap.md
cwperks: In the tests for `LDAPAuthenticationBackend2` it looks like the `type` in the `config.yml` is the fully qualified classname `com.amazon.dlic.auth.ldap2.LDAPAuthenticationBackend2` and I don't see this anywhere in the documentation website. @nibix would you be able to provide more context about the 2 different LDAP authentication backends?

See https://github.com/opensearch-project/security/blob/main/src/test/resources/ldap/config_ldap2.yml#L35

cwperks: Looks like you can also use the short-hand `ldap2` judging by this: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L349-L356 though I don't see any mention of this on the documentation website.

nibix: As far as I remember, the major difference  of the `ldap2` module is indeed connection pooling. Historically, it was introduced as an experimental module in Search Guard which was supposed to replace the `ldap` module at some point in time. Thus, it is also undocumented so far.

cwillum: Changed this to, "OpenSearch can maintain a pool of connections..."

cwillum:
Will leave the v2.4 tag for this PR so that all five connection pooling settings are documented for this version of OS. I'll then create a new PR for v2.3 with only the three connection pooling settings `pool.enabled`, `pool.min_size`,  and `pool.max_size`. Then I'll backport this PR to 2.2, 2.1, 2.0, and 1.3.

PR #1771, opened 2022-10-31
Added instructions for enabling opensearch as a service with systemd

### LINKED ISSUE #1722: [DOC] : Incomplete Documentation for starting the cluster using systemctl command
**What do you want to do?**
 
- [X] Request a change to existing documentation
- [ ] Add new documentation
- [X] Report a technical problem with the documentation
- [ ] Other 

The process of starting the OpenSearch service defined on this URL: https://opensearch.org/docs/latest/opensearch/cluster/#step-5-start-the-cluster is incomplete and misleading. There is no service provided by the tar or another installation method for creating the opensearch.service and user needs to go ahead create it.

So this command:
```
sudo systemctl start opensearch.service
```
gives error with Unit opensearch not found.

https://opensearch.org/docs/latest/opensearch/cluster/#step-5-start-the-cluster

JeffH-AWS:
Signed-off-by: JeffH-AWS <jeffhuss@amazon.com>

### Description
Add a section to the tar installation guide, as an extension of running OS in your own environment, to walk the user through creating a service with systemd. Also added a note to the cluster configuration page so that users who installed OS from a tarball are not blocked from proceeding with cluster configuration.

### Issues Resolved
Fixes #1722 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and subject to the [Developers Certificate of Origin](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/_opensearch/install/tar.md
+++ b/_opensearch/install/tar.md
@@ -453,6 +453,83 @@ $ curl https://your.host.address:9200 -u admin:yournewpassword -k
 }
 ```
 
+### Run OpenSearch as a service with systemd
+
+Create a service for OpenSearch and register it with `systemd`. After the service has been defined, you can enable, start, and stop the OpenSearch process using `systemctl` commands. The following steps reflect an environment where OpenSearch has been installed to `/opt/opensearch`. Change the paths in the following commands and example file to reflect your installation path.
+
+1. Create a user for the OpenSearch service.
+   ```bash
+   sudo adduser --system --shell /bin/bash -U --no-create-home opensearch
+   ```
+
+1. Add your user to the `opensearch` user group.
+   ```bash
+   sudo usermod -aG opensearch $USER
+   ```
+
+1. Change the file owner to `opensearch`. Make sure to change the path if your OpenSearch files are in a different directory.
+   ```bash
+   sudo chown -R opensearch /opt/opensearch/
+   ```
+
+1. Create the service file and open it for editing.
+   ```bash
+   sudo vi /etc/systemd/system/opensearch.service
+   ```
+
+1. Enter the following example service configuration. Make sure to change references to the path if your OpenSearch files are in a different directory.
+   ```bash
+   [Unit]
+   Description=OpenSearch
+   Wants=network-online.target
+   After=network-online.target
+
+   [Service]
+   Type=forking
+   RuntimeDirectory=data
+
+   WorkingDirectory=/opt/opensearch
+   ExecStart=/opt/opensearch/bin/opensearch -d
+
+   User=opensearch
+   Group=opensearch
+   StandardOutput=journal
+   StandardError=inherit
+   LimitNOFILE=65535
+   LimitNPROC=4096
+   LimitAS=infinity
+   LimitFSIZE=infinity
+   TimeoutStopSec=0
+   KillSignal=SIGTERM
+   KillMode=process
+   SendSIGKILL=no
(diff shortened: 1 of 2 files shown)


JeffH-AWS:
Meeting with @peterzhuamazon tomorrow (11/1) for discussion.

JeffH-AWS:
Peter suggested a great idea, which is to include a very clear warning, like what we see [here](https://opensearch.org/artifacts/opensearch/opensearch-2-3-0-min-linux-x64.html). This way, novice users do not potentially cause serious issues while interacting with system components like systemd.

navneet1v:
Thanks for taking this up. This looks good to me

navneet1v (approved):

peterzhuamazon (approved):

_opensearch/install/tar.md
```diff
@@ -453,6 +453,86 @@ $ curl https://your.host.address:9200 -u admin:yournewpassword -k
...
+   KillMode=process
+   SendSIGKILL=no
+   SuccessExitStatus=143
+   TimeoutStartSec=75
+
+   [Install]
+   WantedBy=multi-user.target
+   ```
+
+1. Reload `systemd` manager configuration.
```
ariamarble, line 516: this *might* be a nitpick but I think the periods for this list should be colons?

_opensearch/install/tar.md
```diff
@@ -453,6 +453,86 @@ $ curl https://your.host.address:9200 -u admin:yournewpassword -k
 }
 ```
 
+### Run OpenSearch as a service with systemd
+
+Create a service for OpenSearch and register it with `systemd`. After the service has been defined, you can enable, start, and stop the OpenSearch service using `systemctl` commands. The commands in this section reflect an environment where OpenSearch has been installed to `/opt/opensearch` and should be changed depending on your installation path.
```
ariamarble, line 458: This makes it sound like the user should have added the systemd config *before* starting this section. Maybe something like "To manage OpenSearch as a service, you will need to register it with `systemd`.

ariamarble (approved):
Looks good other than my comments

_opensearch/install/tar.md
JeffH-AWS: I see what you mean. I saw the approvals before the comments so didn't make this change but I'll open a new PR. Good call out.

hdhalter:
@peterzhuamazon - Can you please comment on this? Thanks!

hdhalter:
@peterzhuamazon recommends removing the section altogether. He says that using a manually cooked systemd setup for tar is not recommended because tar is a moving target, and systemd setup is supposed to be stable with no changes, like for deb/rpm. 

@Saikumar282 , thoughts?

hdhalter:
Hi @Saikumar Payavula, thanks for your patience while I rounded up some information for you! Here's some info from our installation developer: If you want a very customizable DIY based on tar, and you also want some service manager to control it without a manual startup using scripts, then yes, systemd is a way, as well as many other similar tools such as supervisord, which is basically the same thing. If you do not care about customization but rather an easier way to set up and start/stop with a built-in service manager support, you can use deb/rpm which comes already bundled with the service scripts for systemd. If you're not sure, you can choose a distribution and try it out. You can read the systemd documentation if you want to manually apply/create a systemd service template rather than using ours bundled in deb/rpm. Hope that helps!

Issue #2155, opened 2022-12-08
[DOC]Revise Security plugin Configuration TOC structure


cwillum:
**What do you want to do?**
 
- [x] Request a change to existing documentation
- [ ] Add new documentation
- [ ] Report a technical problem with the documentation
- [ ] Other 

**Tell us about your request.** 
Reorganize topics in the Security plugin Configuration section to improve logic and order.

**What other resources are available?**
Addresses #1036 for Security.

cwillum:
A rough, first pass at changing this section of the TOC:

Configuration https://opensearch.org/docs/latest/security-plugin/configuration/index/
- Backend configuration (the name of this topic needs more thought) https://opensearch.org/docs/latest/security-plugin/configuration/configuration/
- YAML files https://opensearch.org/docs/latest/security-plugin/configuration/yaml/
- TLS certificates https://opensearch.org/docs/latest/security-plugin/configuration/tls/
- Generate certificates https://opensearch.org/docs/latest/security-plugin/configuration/generate-certificates/
- System indexes (note re-spelling) https://opensearch.org/docs/latest/security-plugin/configuration/system-indices/
- Applying changes (proposed shortening of heading from "Apply changes with the securitytadmin script") https://opensearch.org/docs/latest/security-plugin/configuration/security-admin/
- Configuring sign-in options (previously "Multiple authentication options for Dashboards sign-in") https://opensearch.org/docs/latest/security-plugin/configuration/multi-auth/
- Disable security https://opensearch.org/docs/latest/security-plugin/configuration/disable/

Authentication (suggest Using "Authentication flow" for the index - removing "flow" - and adding a lead-in at the bottom to the next section to introduce specific protocols and options) https://opensearch.org/docs/latest/security-plugin/configuration/concepts/
- Active Directory and LDAP https://opensearch.org/docs/latest/security-plugin/configuration/ldap/
- SAML https://opensearch.org/docs/latest/security-plugin/configuration/saml/
- OpenID Connect https://opensearch.org/docs/latest/security-plugin/configuration/openid-connect/
- Proxy-based authentication https://opensearch.org/docs/latest/security-plugin/configuration/proxy/
-  Client certificate authentication https://opensearch.org/docs/latest/security-plugin/configuration/client-auth/

[...]

PR #2212, opened 2022-12-14
Reorganize Configuration section of the Security TOC

cwillum:

### Description
Break out authentication backend topics in the Configuration section into their own section to clean things up and reorganize in a more logical manner.

### Issues Resolved
Created a new "Authentication backends" section to contain the specific backends that were previously strewn in with straightforward configuration topics.
Also addressing some issues in topics to improve clarity and gaps.
Addresses issue #2155.

### Checklist


cwillum:
@opensearch-project/security Team, I reorganized the TOC for Security documentation. The biggest difference is that I broke out the authentication backends and separated them from general configuration information so they have their own section (more of that to come in a separate issue). @cwperks has seen the reorg. 
My ask is if you'd have a quick look at some language I changed at the intro/top of four sections (in the diffs). Just to be sure it makes sense. I'll list those here. Then you don't have to wade through the many minor TOC changes. Thanks.
_security-plugin/authentication-backends/authc-index.md
_security-plugin/configuration/security-admin.md
_security-plugin/configuration/system-indices.md
_security-plugin/configuration/yaml.md

cwillum:
@Naarcha-AWS Changes to the Security TOC are ready. I'll have you take a look first then pass it to another team member. Big thanks.

_security-plugin/authentication-backends/authc-index.md
```diff
@@ -1,23 +1,28 @@
...
 
-1. To identify a user who wants to access the cluster, the security plugin needs the user's credentials.
+## Authentication flow
+
+1. To identify a user who wants to access the cluster, the Security plugin needs the user's credentials.
 
    These credentials differ depending on how you've configured the plugin. For example, if you use basic authentication, the credentials are a user name and password. If you use a JSON web token, the credentials are stored within the token itself. If you use TLS certificates, the credentials are the distinguished name (DN) of the certificate.
 
-2. The security plugin authenticates the user's credentials against a backend: the internal user database, Lightweight Directory Access Protocol (LDAP), Active Directory, Kerberos, or JSON web tokens.
+2. The Security plugin authenticates the user's credentials against a backend: the internal user database (basic authentictation), Lightweight Directory Access Protocol (LDAP)/Active Directory, JSON web tokens, SAML, or another authentication protocol.
```
cwperks, line 21: Thinking about how to be precise and clear when communication about backend vs. way of communicating auth info (user + password combo or token) to OpenSearch.

There are 2 high-level types of IdP, Identity Providers, backends that Opensearch security plugin interfaces with. 1. Internal User database or 2. an external provider (a backend compliant with LDAP, ODIC or SAML). For client cert, jwt and proxy auth, these could be described as backend-less providers (not completely accurate, but I'll explain) in that OpenSearch does not verify the info passed in the request with a list of users and instead extracts username and backend roles from the request passed to opensearch. 

1. For client cert auth, the principal of the cert is used to extract the username and backend roles. I believe username is the `CN` field and `backend_roles` are the concatenated `OU` fields. As long as the cert is issued by a CA in the `SECURITY_SSL_HTTP_PEMTRUSTEDCAS_FILEPATH` setting the cert will be trusted.
2. For JWT, a token is passed in an HTTP Authorization request header which must contain claims that match the `subject_key` and `roles_key`. The JWT must also be able to be verified with the `signing_key` configured in `config.yml`
3. For proxy auth authentication is performed at the IdP and then forwarded to opensearch. The forwarded request contains the username and backend roles in the HTTP request and OpenSearch trusts the request if its coming from the correct configured source.

The internal database is used with Basic Auth to get username and password info from an HTTP request. So in this case **Basic Auth** is the mechanism for transmitting auth info (username + password) in a request and the **Authentication Backend** would be the internal user database. 

How does this sound?

```
The Security plugin authenticates a request against a configured backend.  Authentication providers include: Basic Auth using the internal user database, Lightweight Directory Access Protocol (LDAP)/Active Directory, JSON web tokens, SAML, or another authentication protocol.
```

This change you are presenting in this PR looks good to me, but stuff to keep in mind.

_security-plugin/authentication-backends/authc-index.md
```diff
@@ -1,23 +1,28 @@
...
 
    The plugin supports chaining backends in `config/opensearch-security/config.yml`. If more than one backend is present, the plugin tries to authenticate the user sequentially against each until one succeeds. A common use case is to combine the internal user database of the security plugin with LDAP/Active Directory.
 
-3. After a backend verifies the user's credentials, the plugin collects any backend roles. These roles can be arbitrary strings in the internal user database, but in most cases, these backend roles come from LDAP/Active Directory.
+3. After a backend verifies the user's credentials, the plugin collects any backend roles. These roles can be arbitrary strings in the internal user database, roles retrieved from the LDAP/Active Directory server, or roles that are kept as attributes with the SAML protocol.
```
cwperks, line 25: Extraction of backend roles is applicable for all backends except for basic auth with the internal user database. LDAP is a bit exceptional because `authz` of the security plugin's `config.yml` needs to be configured to specify how roles should be extracted from the directory service. 

For the internal user database, roles are determined by roles that the user is mapped to within opensearch. With an external IdP, a cluster admin must map backend roles to roles in OpenSearch and users will transitively get OpenSearch roles based on the roles mapping. i.e. In LDAP I may be in the `My Org Developers` group, and in opensearch there could be a role named `developers` which has a certain set of permissions. The cluster admin would need to map `My Org Developers` to the `developers` role to map the external group to the internal role within OpenSearch.

cwperks (commented):
Thank you @cwillum. I like breaking up the different authentication backends into menu items. I think that helps with discovery and navigation on the documentation website.

I left a couple of comments that I think can be addressed in future PRs as we work to make the documentation more precise and easy-to-follow.

_security-plugin/access-control/api.md
```diff
@@ -678,11 +678,11 @@ PUT _plugins/_security/api/roles/<role>
 }
 ```
 
->Due to word boundaries associated with Unicode special characters, the Unicode standard analyzer cannot index a [text field type](https://opensearch.org/docs/2.2/opensearch/supported-field-types/text/) value as a whole value when it includes one of these special characters. As a result, a text field value that includes a special character is parsed by the standard analyzer as multiple values separated by the special character, effectively tokenizing the different elements on either side of it.
+>Due to word boundaries associated with Unicode special characters, the Unicode standard analyzer cannot index a [text field type]({{site.url}}{{site.baseurl}}/opensearch/supported-field-types/text/) value as a whole value when it includes one of these special characters. As a result, a text field value that includes a special character is parsed by the standard analyzer as multiple values separated by the special character, effectively tokenizing the different elements on either side of it.
```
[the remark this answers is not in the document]
cwillum: I believe we're using an automated link checker now. Not sure if it would identify this specific problem. But will look into it.

_security-plugin/access-control/field-level-security.md
```diff
@@ -93,7 +93,7 @@ someonerole:
 
 ### REST API
 
-See [Create role]({{site.url}}{{site.baseurl}}/security-plugin/access-control/api#create-role).
+See [Create role]({{site.url}}{{site.baseurl}}/security-plugin/access-control/api/#create-role).
```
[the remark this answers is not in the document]
cwillum: Apparently both ways work. I just followed the lead of the web server, which adds the forward slash when you use the link. The doc team informed me that behavior originates with the web server. In any case, I see now there's no need bothering to add it from now on. Thanks.

_security-plugin/authentication-backends/authc-index.md
cwillum: I see the distinction. And those are good points to keep in mind. I tweaked the suggestion a little. Does it still make sense, or have I taken it off course?

`The Security plugin authenticates a server request against a backend configured for an authentication provider.  Some examples of authentication providers used with OpenSearch include Basic Auth (which uses the internal user database), Lightweight Directory Access Protocol (LDAP)/Active Directory, JSON web tokens, SAML, or another authentication protocol.`

_security-plugin/authentication-backends/authc-index.md
cwillum: @cwperks I tried to sum this up while keeping it general.
`After a backend verifies the user's credentials, the plugin collects any [backend roles]({{site.url}}{{site.baseurl}}/security-plugin/access-control/index/#concepts). The authentication provider determines the way these roles are retrieved. For example, LDAP extracts backend roles from its directory service based on their mappings to roles in OpenSearch, while SAML stores the roles as attributes. When basic authentication is used, the internal user database refers to role mappings configured in OpenSearch.`

cwperks: Sounds good to me

_security-plugin/authentication-backends/authc-index.md
cwperks: @cwillum The tweaked suggestion sounds good to me, but instead of `server request` can it simply be `request`?

Judging by item 1, the word Credentials is used generically to mean what's needed to confer access so its accurate to say that credentials must be included on every request. 

For most of the backends, Username + Password will be used for the very first request so its accurate to say the credentials are included in the request. In many cases, subsequent requests will use an access token (JWT) to prevent the need of passing username + password on every request.

In item 1:

> the credentials are stored within the token itself

When I first read this it sounded like password info was embedded in the token, but that's not the case. The username (`subject_key`) and roles (`roles_key`) are embedded in the token.

cwillum: RE: "... but instead of server request can it simply be request?"
Definitely. I'm probably unnecessarily splitting hairs with some of these proposed changes to wording. But it's helpful to refine along the way. Thanks for your patience.

cwillum:
Still waiting for a review on this from @Naarcha-AWS . Removed Tech Review tag and replaced with Doc Review. Will sound out a request on doc team Slack channel once the first is done.

_security-plugin/access-control/permissions.md
```diff
@@ -41,7 +41,7 @@ These permissions also allow you add, update, or delete documents (e.g. `PUT tes
 
 ## Test permissions
 
-If you want a user to have the absolute minimum set of permissions necessary to perform some function---the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege)----the best way is to send representative requests to your cluster as a new test user. In the case of a permissions error, the security plugin is very explicit about which permissions are missing. Consider this request and response:
+If you want a user to have the absolute minimum set of permissions necessary to perform some function—the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege)—the best way is to send representative requests to your cluster as a new test user. In the case of a permissions error, the security plugin is very explicit about which permissions are missing. Consider this request and response:
```
DarshitChanpura, line 44: Should p be capital as in `[Principle of least privilege]`

_security-plugin/configuration/system-indices.md
```diff
@@ -1,25 +1,26 @@
...
 plugins.security.system_indices.enabled: true
 plugins.security.system_indices.indices: [".opendistro-alerting-config", ".opendistro-alerting-alert*", ".opendistro-anomaly-results*", ".opendistro-anomaly-detector*", ".opendistro-anomaly-checkpoints", ".opendistro-anomaly-detection-state", ".opendistro-reports-*", ".opendistro-notifications-*", ".opendistro-notebooks", ".opendistro-asynchronous-search-response*"]
 ```
 
-To access these indices, you must authenticate with an [admin certificate]({{site.url}}{{site.baseurl}}/security-plugin/configuration/tls#configure-admin-certificates):
-
-```bash
-curl -k --cert ./kirk.pem --key ./kirk-key.pem -XGET 'https://localhost:9200/.opendistro_security/_search'
-```
+You can add additional system indexes in `opensearch.yml`. An alternative way to remove indexes is to delete the index from the `plugins.security.system_indices.indices` list on each node and restart OpenSearch.
```
DarshitChanpura, line 25: nit: `indexes` -> `indices`

And we can reword: "An alternative way to remove system indices is to delete them `plugins.security.system_indices.indices\` list on each node and restart OpenSearch". 

Also, I'm not sure what the original way to remove is. Should it be documented here ( maybe I'm missing something here). Thoughts?

_security-plugin/configuration/yaml.md
```diff
@@ -1,15 +1,17 @@
...
 
-The best use of these YAML files is to configure [reserved and hidden resources]({{site.url}}{{site.baseurl}}/security-plugin/access-control/api#reserved-and-hidden-resources), such as the `admin` and `kibanaserver` users. You might find it easier to create other users, roles, mappings, action groups, and tenants using OpenSearch Dashboards or the REST API.
+Before running [`securityadmin.sh`]({{site.url}}{{site.baseurl}}/security-plugin/configuration/security-admin/) to load the settings into the `.opendistro_security` index, perform an initial configuration of the YAML files. The files can be found in the `config/opensearch-security` directory. It's also good practice to back up these files so that you can reuse them for other clusters.
+
+One approach to using the YAML files—which we recommend—is to first configure [reserved and hidden resources]({{site.url}}{{site.baseurl}}/security-plugin/access-control/api#reserved-and-hidden-resources), such as the `admin` and `kibanaserver` users. Thereafter you can create other users, roles, mappings, action groups, and tenants using OpenSearch Dashboards or the REST API.
```
DarshitChanpura, line 14: Thoughts on using `-` here instead of `,` surrounding `which we recommend`?

_security-plugin/authentication-backends/authc-index.md
```diff
@@ -0,0 +1,31 @@
...
+nav_order: 45
+has_children: true
+has_toc: false
+redirect_from:
+  - /security-plugin/authentication-backends/
+---
+
+# Authentication backends
+
+Authentication backend configurations determine the method or methods you use for authenticating users and the way users pass their credentials and sign in to OpenSearch. Having an understanding of the basic authentication flow before getting started can help with the configuration process for whichever backend you choose. Consider the high-level sequence of events in the description that follows, and then refer to the detailed steps for configuring the authentication type you choose to use with OpenSearch.
```
DarshitChanpura, line 13: Thoughts on: `.. the method/s to be used for authenticating users ..`

_security-plugin/access-control/api.md
```diff
@@ -2,7 +2,7 @@
 layout: default
 title: API
 parent: Access control
-nav_order: 90
+nav_order: 120
```
DarshitChanpura, line 5: Generic question: Changing `nav_order` will re-organize this document based on the value entered. Correct?

_security-plugin/configuration/system-indices.md
DarshitChanpura: Re: `An alternative way..`. This should definitely be addressed at some point to clearly defined in the document. Not urgent for this change tho.

Re: `indices/indexes` I've been confused about this too, but as you mentioned in the dev world afaik it has always been indices. However, I'm completely fine with either as long as the whole documentation is aligned and points to the same usage

Re: `\` yes the backslash was a typo on my end. It shouldn't contain any backslash. And, I like your suggestion to rewrite it to be singular.

_security-plugin/configuration/yaml.md
DarshitChanpura: yes this is even better

_security-plugin/authentication-backends/authc-index.md
DarshitChanpura: Yes, singular usage sounds good and it won't divert the user interpretation of this sentence.

_security-plugin/access-control/api.md
DarshitChanpura: oo nice..thank you for sharing the final view. This helps a lot in visualizing the "re-organization" of the Security Plugin  section

cwillum: Agreed. It helps.

DarshitChanpura (approved):
Thanks @cwillum !

Naarcha-AWS (approved):

hdhalter (approved):
This looks good! Thanks, Chris and Nate.

PR #3266, opened 2023-03-07
Document class metadata memory reservation impacts

dbwiddis:
### Description

The `bootstrap.memory_lock=true` imposes a hard lower limit on native memory reservation by the JVM, which include 1 GB for class metadata in addition to heap space.

On small VMs (such as t1.micro) with only 1 GB space, this results in out of memory errors.

This change updates the documentation to alert users to this setting.

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch/issues/5865

### Checklist

### DIFF
--- a/_install-and-configure/install-opensearch/index.md
+++ b/_install-and-configure/install-opensearch/index.md
@@ -71,7 +71,7 @@ The [sample docker-compose.yml]({{site.url}}{{site.baseurl}}/install-and-configu
 
   Disables swapping (along with `memlock`). Swapping can dramatically decrease performance and stability, so you should ensure it is disabled on production clusters.
 
-  Enabling the `bootstrap.memory_lock` setting will cause the JVM to reserve any memory it needs. On VMs with limited memory (such as t2.micro with 1GB) the [default 1GB Class Metadata](https://docs.oracle.com/javase/10/gctuning/other-considerations.htm#JSGCT-GUID-BFB89453-60C0-42AC-81CA-87D59B0ACE2E) native memory reservation result in error due to the lack of native memory. To prevent errors, limit the reserved memory size using `-XX:CompressedClassSpaceSize` or `-XX:MaxMetaspaceSize` and set the size of the Java heap to make sure you have enough system memory.
+  Enabling the `bootstrap.memory_lock` setting will cause the JVM to reserve any memory it needs. On VMs with limited memory, the [default 1 gigabyte (GB) Class Metadata](https://docs.oracle.com/javase/10/gctuning/other-considerations.htm#JSGCT-GUID-BFB89453-60C0-42AC-81CA-87D59B0ACE2E) native memory reservation may result in an error due to the lack of native memory. To prevent errors, limit the reserved memory size using `-XX:CompressedClassSpaceSize` or `-XX:MaxMetaspaceSize` and set the size of the Java heap to make sure you have enough system memory.
 
 - `OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m`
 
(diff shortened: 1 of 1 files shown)


Naarcha-AWS (approved):

_install-and-configure/install-opensearch/index.md
```diff
@@ -71,6 +71,8 @@ The [sample docker-compose.yml]({{site.url}}{{site.baseurl}}/install-and-configu
 
   Disables swapping (along with `memlock`). Swapping can dramatically decrease performance and stability, so you should ensure it is disabled on production clusters.
 
+  Enabling the `bootstrap.memory_lock` setting will cause the JVM to reserve any memory it needs. On VMs with limited memory (such as t2.micro with 1GB) the [default 1GB Class Metadata](https://docs.oracle.com/javase/10/gctuning/other-considerations.htm#JSGCT-GUID-BFB89453-60C0-42AC-81CA-87D59B0ACE2E) native memory reservation result in error due to the lack of native memory. To prevent errors, limit the reserved memory size using `-XX:CompressedClassSpaceSize` or `-XX:MaxMetaspaceSize` and set the size of the Java heap to make sure you have enough system memory.
```
JeffH-AWS, line 74: Please remove the reference to `t2.micro` as that's specific to Amazon and this documentation should be platform agnostic.

For the link (i.e., `[default 1GB Class Metadata](https://docs.oracle.com/javase/10/gctuning/other-considerations.htm#JSGCT-GUID-BFB89453-60C0-42AC-81CA-87D59B0ACE2E)`) I'm not sure, as a reader, where on that page I'm supposed to look. Ctrl+F looking for `1 GB` and `1GB` on that page yield no results. Please add context to the link so the reader knows why they're being sent to the other page.

Formal cross-references should also, ideally, match the title of the page you link to.

Typo: `native memory reservation result in error due to the lack of native memory.` Singular "reservation" or wrong conjugated form of the verb "result." Should be "reservations result" or "reservation results."

Also, when describing a unit of measure make sure to add a space between the number and the unit. So `1GB` should be written as `1 GB`.

JeffH-AWS (changes requested):

_install-and-configure/install-opensearch/index.md
dbwiddis: Thanks for the review, @JeffH-AWS .  I've removed the AWS-specific reference and fixed the grammar.  I also changed the link to include terms that would show up in a search on that page, specifically "1 gigabyte (GB)".  

I did include "Class Metadata" in the link as well, which is the section on that page relevant to this discussion; and in fact someone tuning memory on a small VM would probably want to read that whole section.  

I could not find any other way to more directly reference that content. [This link](https://docs.oracle.com/javase/9/gctuning/other-considerations.htm#JSGCT-GUID-B29C9153-3530-4C15-9154-E74F44E3DAD9:~:text=The%20size%20of%20the%20region%20can%20be%20set%20with%20%2DXX%3ACompressedClassSpaceSize%20and%20is%201%20gigabyte%20(GB)%20by%20default) would work on Chrome, but not other browsers.

Let me know if I should change this to something like this:

Enabling the `bootstrap.memory_lock` setting will cause the JVM to reserve any memory it needs. The [Java Platform, Standard Edition HotSpot Virtual Machine Garbage Collection Tuning Guide section on Class Metadata](https://docs.oracle.com/javase/9/gctuning/other-considerations.htm#JSGCT-GUID-B29C9153-3530-4C15-9154-E74F44E3DAD9) documents a default 1 gigabyte (GB) Class Metadata native memory reservation. Combined with Java heap, this may result in an error due to the lack of native memory on VMs with less memory than these requirements. To prevent errors, limit the reserved memory size using `-XX:CompressedClassSpaceSize` or `-XX:MaxMetaspaceSize` and set the size of the Java heap to make sure you have enough system memory.

Alternate less verbose title:  "Java SE Hotspot VM Garbage Collection Tuning Guide"

JeffH-AWS: That sounds great! Very nice work thanks so much Daniel!

kolchfa-aws (approved):

PR #3426, opened 2023-03-11
Add documentation to rename opensearch_dashboards predefined roles to kibana

### LINKED ISSUE #2723: [DOC] Correct naming of predefined `opensearch_dashboards_readonly` role and `opensearch_dashboards_user` user
**What do you want to do?**
 
- [ ] Add new documentation
- [ ] Other 

 Coming from https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3391.
From our documentation,[predefined roles](https://opensearch.org/docs/2.4/security/access-control/users-roles/#predefined-roles) and the example below is using `opensearch_dashboards_read_only` and `openesarch_dashboards_user` as the name of user and role. But actually from the security plugin [codebase](https://github.com/opensearch-project/security/blob/75b6c3055f99ed8df0fdbeb10756a865cf3e6c76/config/roles.yml#L6), you can see it's named `kibana_read_only`. Similar with `opensearch_dashboards_user` --> `kibana_user`

Please correct the naming on our documentation to avoid misleading users.

cwillum:
### Description
Somehow the predefined roles for providing permissions for all functionality in Dashboards and read-only permissions became named `openesarch_dashboards_user` and `opensearch_dashboards_read_only` while in the code they remain `kibana_user` and `kibana_read_only`. Need to name them properly to agree with the code.

### Issues Resolved
renamed these roles `kibana_user` and `kibana_read_only` in the [Predefined roles](https://opensearch.org/docs/latest/security/access-control/users-roles/#predefined-roles) table in Users and Roles documentation.

Fixes #2723.

### Checklist

### DIFF
--- a/_security/access-control/users-roles.md
+++ b/_security/access-control/users-roles.md
@@ -113,8 +113,8 @@ Role | Description
 `cross_cluster_replication_leader_full_access` | Grants full access to perform cross-cluster replication actions on the leader cluster.
 `observability_full_access` | Grants full access to perform actions on Observability objects such as visualizations, notebooks, and operational panels.
 `observability_read_access` | Grants permission to view Observability objects such as visualizations, notebooks, and operational panels, but not create, modify, or delete them.
-`opensearch_dashboards_read_only` | A special role that prevents users from making changes to visualizations, dashboards, and other OpenSearch Dashboards objects. See `opensearch_security.readonly_mode.roles` in `opensearch_dashboards.yml`. Pair with the `opensearch_dashboards_user` role.
-`opensearch_dashboards_user` | Grants permissions to use OpenSearch Dashboards: cluster-wide searches, index monitoring, and write to various OpenSearch Dashboards indices.
+`kibana_read_only` | A special role that prevents users from making changes to visualizations, dashboards, and other OpenSearch Dashboards objects. See `opensearch_security.readonly_mode.roles` in `opensearch_dashboards.yml`. Pair with the `kibana_user` role.
+`kibana_user` | Grants permissions to use OpenSearch Dashboards: cluster-wide searches, index monitoring, and write to various OpenSearch Dashboards indices.
 `logstash` | Grants permissions for Logstash to interact with the cluster: cluster-wide searches, cluster monitoring, and write to the various Logstash indices.
 `manage_snapshots` | Grants permissions to manage snapshot repositories, take snapshots, and restore snapshots.
 `readall` | Grants permissions for cluster-wide searches like `msearch` and search permissions for all indices.


cwillum:
@opensearch-project/opensearch-dashboards-core It looks like the role naming issue began in [#3391](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3391). This PR renames the roles so they agree with the predefined roles in Security code. 
But in the definition for what is now the `kibana_read_only` role, it says, "A special role that prevents users from making changes to visualizations, dashboards, and other OpenSearch Dashboards objects." And then, "See `opensearch_security.readonly_mode.roles` in `opensearch_dashboards.yml`." We don't have a reference in documentation that shows the `opensearch_dashboards.yml` file with all of its properties/settings defined, which a user could look at to find out more about this.
What is the relationship between the `kibana_read_only` role and the `opensearch_security.readonly_mode.roles` setting that you add to `opensearch_dashboards.yml`? Are there dependencies? Do they do different things? Is there a file in the OSD repo that defines this setting? Thanks.

joshuarrrr:
Just wanted to mention that the source file for `opensearch_dashboards.yml` actually does have example settings and values, all commented out: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/config/opensearch_dashboards.yml

The problem is that it only lists configuration settings that are present in the core dashboards application, so doesn't include config settings defined by other plugins (such as the security-dashboards plugin, in this case).

cwillum:
@joshuarrrr Thanks for the clarification. It looks like this is a question for security dashboards.
@opensearch-project/security Does anyone have information on this `opensearch_security.readonly_mode.roles` setting that you add in the `opensearch_dashboards.yml`? I'm hoping to clarify information in the `kibana_read_only` role definition in docs: [Predefined roles](https://opensearch.org/docs/latest/security/access-control/users-roles/#predefined-roles). (It's still called `opensearch_dashboards_read_only` here).

DarshitChanpura:
There names in security plugin are still `kibana_read_only` and `kibana_user` instead of `opensearch_dashboards_read_only` and `opensearch_dashboards_user`. 
Renaming these to `opensearch_dashboards_*` is a significant and a breaking change. @opensearch-project/security Should we go ahead and make a PR to rename these?

To answer your question @cwillum, `opensearch_security.readonly_mode.roles` setting marks the roles passed in it as read only. The roles added here are read by security plugin via this piece of code: https://github.com/opensearch-project/security-dashboards-plugin/blob/main/public/plugin.ts#L88-L94.

DarshitChanpura:
`opensearch_security.readonly_mode.roles` setting is used to activate the read-only mode. Roles that are defined as dashboards "read-only" should be added to this setting for them to be effective:

So in this case, if as a user is assigned `opensearch_dashboards_read_only` role but this role has not been added to `opensearch_security.readonly_mode.roles` setting then it won't work.
For dashboards read-only mode to work for a user, they need to be assigned a read-only role and that role must be present in the `opensearch_security.readonly_mode.roles` setting in `opensearch_dashboards.yml`?

Hope this adds some clarification to answer your question

cwillum:
@DarshitChanpura That's a great explanation and super helpful. I wouldn't have known. Do know if multiple roles are listed as an array, separated by a comma?

JeffH-AWS (approved):

DarshitChanpura:
>  Do you know if multiple roles are listed as an array, separated by a comma?

That is correct.

cwillum:
Waiting on availability for doc team review.

PR #3649, opened 2023-04-01
Adding documentation for behavior of indices.recovery.max_bytes_per_sec set to `0mb` or less

### LINKED ISSUE #3648: [DOC] missing documentation for setting "indices.recovery.max_bytes_per_sec" to <= 0
**What do you want to do?**
 
- [ ] Add new documentation
- [ ] Other 

 the "indices.recovery.max_bytes_per_sec" cluster setting has [unexpected behavior for values less than or equal to 0](https://github.com/opensearch-project/OpenSearch/blob/c85d33e2e487242eb5ad60b64567ca2ec14fc88f/server/src/main/java/org/opensearch/indices/recovery/RecoverySettings.java#L256-L265).
 
 This causes the rate limiting to be eliminated, for a value of `0.0mb` I would expect rate limiting to be constant - as in - no data is being recovered as long as the value is 0. However the opposite happens, and there is no rate limiting, so data transfers at the highest rate possible. 

I suggest changing the docs to reflect this behavior.
 
setup a two node cluster.
add a 100 byte doc to a **single-sharded index**.
set the param "indices.recovery.max_bytes_per_sec" to "0.000001mb" (1 byte per sec)
exclude the node that has the index on it. 
after at least 100 seconds the index is moved to the other node, and not much faster than this.

repeat the test, except set "indices.recovery.max_bytes_per_sec" to "0mb" 
the index moves very quickly in this case after excluding the node.

AmiStrn:
After a lovely surprise when trying to limit the recovery completely and discovering the opposite was happening I see we need the documentation to include this nice extra info.

### Description
missing documentation for setting `indices.recovery.max_bytes_per_sec` to `0mb` or less.

### Issues Resolved
Resolves #3648

### Checklist

### DIFF
--- a/_api-reference/cluster-api/cluster-settings.md
+++ b/_api-reference/cluster-api/cluster-settings.md
@@ -63,7 +63,7 @@ The following request field parameters are compatible with the cluster API.
 | action.auto_create_index | Boolean | Automatically creates an index if the index doesn't already exist. Also applies any index templates that are configured. Default is `true`. |
 | action.destructive_requires_name | Boolean | When set to `true`, you must specify the index name to delete an index. You cannot delete all indexes or use wildcards. Default is `true`. |
 | cluster.indices.close.enable | Boolean | Enables closing of open indexes in OpenSearch. Default is `true`. |
-| indices.recovery.max_bytes_per_sec | String | Limits the total inbound and outbound recovery traffic for each node. This applies to peer recoveries and snapshot recoveries. Default is `40mb`. Setting to values less than or equal to `0mb` will result in disabling this rate limiting. |
+| indices.recovery.max_bytes_per_sec | String | Limits the total inbound and outbound recovery traffic for each node. This applies to peer recoveries and snapshot recoveries. Default is `40mb`. If you set the recovery traffic value to less than or equal to `0mb`, rate limiting will be disabled, which causes recovery data to be transferred at the highest possible rate. |
 | indices.recovery.max_concurrent_file_chunks | Integer | The number of file chunks sent in parallel for each recovery operation. Default is `2`. |
 | indices.recovery.max_concurrent_operations | Integer | The number of operations sent in parallel for each recovery. Default is `1`. |
 | logger.org.opensearch.discovery | String | Loggers accept Log4j2’s built-in log levels: `OFF`, `FATAL`, `ERROR`, `WARN`, `INFO`, `DEBUG`, and `TRACE`. Default is `INFO`. |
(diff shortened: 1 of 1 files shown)


AmiStrn:
@hdhalter @Naarcha-AWS @dtaivpp who reviews PRs for this repo?

AmiStrn:
@nknize do you think this can be categorized as a bug in the core project? 
The behavior was surprising, if you set tis param to 0mb per second then the behavior is basically "as many mb as possible per second"...
I opened this so it is at least documented somewhere.

Naarcha-AWS:
> @hdhalter @Naarcha-AWS @dtaivpp who reviews PRs for this repo?

@AmiStrn: That would be anyone on the documentation team listed [here](https://github.com/opensearch-project/documentation-website#points-of-contact). I'm taking a look now.

_api-reference/cluster-api/cluster-settings.md
```diff
@@ -63,7 +63,7 @@ The following request field parameters are compatible with the cluster API.
 | action.auto_create_index | Boolean | Automatically creates an index if the index doesn't already exist. Also applies any index templates that are configured. Default is `true`. |
 | action.destructive_requires_name | Boolean | When set to `true`, you must specify the index name to delete an index. You cannot delete all indexes or use wildcards. Default is `true`. |
 | cluster.indices.close.enable | Boolean | Enables closing of open indexes in OpenSearch. Default is `true`. |
-| indices.recovery.max_bytes_per_sec | String | Limits the total inbound and outbound recovery traffic for each node. This applies to peer recoveries and snapshot recoveries. Default is `40mb`. |
+| indices.recovery.max_bytes_per_sec | String | Limits the total inbound and outbound recovery traffic for each node. This applies to peer recoveries and snapshot recoveries. Default is `40mb`. Setting to values less than or equal to `0mb` will result in disabling this rate limiting. |
```
Naarcha-AWS, line 66:
```suggestion
| indices.recovery.max_bytes_per_sec | String | Limits the total inbound and outbound recovery traffic for each node. This applies to peer recoveries and snapshot recoveries. Default is `40mb`. If you set the recovery traffic  value to less than or equal to `0mb`, rate limiting will be disabled, which causes recovery data to be transferred at the highest possible rate. |
```

Naarcha-AWS (changes requested):
Adding additional context based on your issue. Feel free to tweak as needed.

_api-reference/cluster-api/cluster-settings.md
AmiStrn: I'll make a commit via my fork. If i commit from here it will cause dco issues. 
The change you propose is acceptable.

AmiStrn:
> > @hdhalter @Naarcha-AWS @dtaivpp who reviews PRs for this repo?
> 
> 
> 
> @AmiStrn: That would be anyone on the documentation team listed [here](https://github.com/opensearch-project/documentation-website#points-of-contact). I'm taking a look now.
> 
> 

Thanks for taking a look @Naarcha-AWS

PR #3898, opened 2023-04-27
Revert "[Backport 2.3] Add new settings for SAML and OIDC that allow for cookie splitting"

cwillum:
Reverts opensearch-project/documentation-website#3888. This is only released in 2.7.


cwillum:
This is only released in 2.7.

PR #4411, opened 2023-06-28
Add documentation for on-behalf-of token authorization

### LINKED ISSUE #4388: [DOC]  Add documentation for OnBehalfOf Authentication
**What do you want to do?**
 
- [ ] ~Request a change to existing documentation~
- [x] Add new documentation
- [ ] ~Report a technical problem with the documentation~
- [ ] ~Other~

### Description of the Feature 
the security team has opted to develop an on-behalf-of Json web token for managing authentication requests between the user's client and the extension just in time. For the initial extensions design, these tokens will allow the extension to interact with the OpenSearch cluster using the same privileges as the initiating user, that’s the reason it is named as “on-behalf-of”. However, since this token doesn’t have any restrictions in place, it also has the capability of letting a service operate as if they are the original user for some window of time. Which means this generic feature is not limited extension related use cases..

For the most of the scenarios, this obo token will be used just-in-time which means that the token is issued just before it is needed for usage. The token will have a configurable short time window of validity after which it is expired and no longer good for use. If the extension REST handler finished quickly, the token may still be valid after the request is done. If the extension REST handler takes a long time then there is a risk that the token will expire before the request has been fully handled. There may need to be a mechanism in place to refresh a token or have certain handlers be able to obtain tokens with a longer window of validity, for example I think 5 minutes is reasonable and most handlers should be able to handle a request in a short period of time.

#### Generic Usage vs. Extension Project ###
Just-in-time UsageFor extensions project, the security team has opted to develop an on-behalf-of Json web token for managing authentication requests between the user's client and the extension just in time. For the initial extensions design, these tokens will allow the extension to interact with the OpenSearch cluster using the same privileges as the initiating user, that’s the reason it is named as “on-behalf-of”. However, since this token doesn’t have any restrictions in place, it also has the capability of letting a service operate as if they are the original user for some window of time. Which means this generic feature is not limited extension related use cases..

For the most of the scenarios, this obo token will be used just-in-time which means that the token is issued just before it is needed for usage. The token will have a configurable short time window of validity after which it is expired and no longer good for use. If the extension REST handler finished quickly, the token may still be valid after the request is done. If the extension REST handler takes a long time then there is a risk that the token will expire before the request has been fully handled. There may need to be a mechanism in place to refresh a token or have certain handlers be able to obtain tokens with a longer window of validity, for example I think 5 minutes is reasonable and most handlers should be able to handle a request in a short period of time.

### Token’s Payload
```
Issuer (iss): OpenSearch Cluster Identifier
Issue-at (iat): Current time of issuing this token
Expiry (exp): Expiration time
Subject (sub): User ID
Audience (aud): Extension’s unique identifier
Roles: Security Privilege Evaluation
```
### Roles in OBO's Payload ###
Plugin Backward Compatibility Mode (BWC Mode)
```
BWC Mode OFF (default):
Encrypted roles (er)
BWC Mode ON:
Decrypted roles in plain-text (dr)
Decrypted backend roles in plain-text (dbr)
```
### Configuration in `Config.yml`
```
on_behalf_of:
    signing_key: xxxxxxxxxx
    encryption_key: xxxxxxxx
```

### References ###
- https://github.com/opensearch-project/security/issues/2573
- https://github.com/opensearch-project/security/issues/2865

cwillum:
### Description
On-behalf-of tokens will allow users to authorize services running with OpenSearch on their behalf. The first use case for this token is the Security extension.

### Issues Resolved
Fixes #4388 

### Checklist


cwillum:
Pausing this for 2.9. If it doesn't get fresh attention between release cycles, it will be an issue for 2.10.

cwillum:
Pushed to 2.10.

cwillum:
@RyanL1997 I'm getting near to completing a rough draft for OBO authentication. A few questions remain:
- Has anything changed in configuration/usage since the documentation issue #4388 was created?
- You can see I've created some questions in comments (inline with text in "Files changed"). Basically they have to do with `authc` configuration in the `config.yml` file (see changes). The JWT auth domain doesn't appear to require an `encryption_key` setting in its configuration. Is there a special use here for OBO? Are there additional configuration settings for OBO that carry over from a JWT domain configuration (e.g., challenge to false, the `jwt_header`, set authentication_backend type to `noop`)?
- I'm not quite clear on the "Roles in OBO's Payload" section in the documentation issue. What is this about? Where is this configured? Is there more information somewhere about the part this plays in OBO and how it's used?

Thanks.

PR #4726, opened 2023-08-09
Removed forced options, cleaned up link checker args.

dblock:
### Description

It seems like the `@check_forced_external` option was added to enable checking internal links marked as external e.g. `/docs`, however the forced external matcher `@forced_external_matcher = %r{^https?://.*(?=opensearch\.org/)}.freeze` seems to just say "anything on opensearch.org". Jekyll renders all URLs using the base URL, so everything would match that now. I don't think this option is useful in its current state at all. 

@AMoo-Miki correct me if I am wrong.

This PR also makes passing arguments a bit easier to read, removing the magical `[1..3]` check in incremental options in favor of using `Ruby::Enum`, and renames `should_build_fatally` to `fail_on_error` which is more intentional.

### Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,3 +39,4 @@ gem "webrick", "~> 1.7"
 # Link checker
 gem "typhoeus"
 gem "ruby-link-checker"
+gem "ruby-enum"
--- a/_plugins/link-checker.rb
+++ b/_plugins/link-checker.rb
@@ -12,12 +12,13 @@ require 'uri'
 require 'pathname'
 require 'typhoeus'
 require 'ruby-link-checker'
+require 'ruby-enum'
 
 ##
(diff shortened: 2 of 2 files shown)


kolchfa-aws (approved):
Thanks, @dblock!

AMoo-Miki:
Forced matcher was meant to treat links on the website and forum as external to the docsite despite them sharing a domain name. The code has changed a lot and I am not sure how something like `https://opensearch.org/bad/url` is validated now.

dblock:
> Forced matcher was meant to treat links to the website and forum as external to the docsite despite them sharing a domain name. The code has changed a lot and I am not sure how something like `https://opensearch.org/bad/url` is validated now.

Looks like there are no such links left in this project (or link checker would have failed). If needed we can re-add a list of URLs to ignore, feels simpler.

PR #5581, opened 2023-11-13
Update Security Docs: Clarify Anomaly Detection Access Control

kaituo:
### Description
This commit updates the security documentation to clarify how `anomaly_read_access` and `anomaly_full_access` permissions impact user access to anomaly detection results. It explains that users without backend roles can still view each other's detection results with these permissions. The documentation now also emphasizes the importance of using backend role filters for more granular access control to detector results.

### Checklist


_observing-your-data/ad/security.md
```diff
@@ -92,3 +92,5 @@ PUT _plugins/_security/api/rolesmapping/anomaly_full_access
 ```
 
 Because they have different backend roles, `alice` and `bob` cannot view each other's detectors or their results.
+
+If users lack backend roles, they can still view each other's anomaly detection results if they possess anomaly_read_access. This is also true for anomaly_full_access, as it includes all the permissions of anomaly_read_access. Administrators should inform users that having anomaly_read_access permits viewing results from any detector in the cluster, potentially including data not directly accessible to them. To limit access to detector results more specifically, administrators should use backend role filters during detector creation. This way, only users with matching backend roles can access results from those specific detectors.
```
vagimeli, line 96:
```suggestion
If users do not have backend roles, they still can view other users' anomaly detection results as long as they have `anomaly_read_access`. This is the same for users who have `anomaly_full_access`, as it includes all of the permissions as `anomaly_read_access`. Administrators should inform users that having `anomaly_read_access` allows for viewing of the results from any detector in the cluster, including data not directly accessible to them. To limit access to the detector results, administrators should use backend role filters at the time the detector is created. This ensures only users with matching backend roles can access results from those particular detectors.
```

vagimeli (approved):
Editorial review complete. I made some edits for clarity.

PR #5821, opened 2023-12-11
Add allocation.require.temp config in ism example guide

10000-ki:
### Description

https://opensearch.org/docs/latest/im-plugin/ism/policies/#example-policy

If we follow the sample policy settings, the index will not move to the warm node.

The `allocation.require.temp` setting must be included as an action

### Issues Resolved
None

### Checklist


andrross:
Thanks @10000-ki for the contribution! This change makes sense, but my concern is that the existing description of the policy is:

> In this case, an index is initially in a hot state. After a day, it changes to a warm state, where the number of replicas increases to 5 to improve the read performance.
> 
> After 30 days, the policy moves this index into a delete state. The service sends a notification to a Chime room that the index is being deleted, and then permanently deletes it.

This defines "warm" as being a state with increased replicas but doesn't mention anything about the cluster setup of configuring "warm" nodes. My concern is that adding the allocation requirement might be confusing. If we add the allocation requirement, then I think we'll need some more details about cluster setup the explains how and why nodes are assigned the "warm" attribute. What do you think?

10000-ki:
>  If we add the allocation requirement, then I think we'll need some more details about cluster setup the explains how and why nodes are assigned the "warm" attribute. What do you think?

good

`node.attr.temp` / `allocation.require.temp`

If we explain above config, I think users will like it

스크린샷 2023-12-22 오후 3 28 51

most people would think that the index moves from the hot node to the warm node if they look at the picture above.
but I think this is a picture of a simple 'state' change.
so there may be a little confusion.

@andrross 

so

Why don't you add it under the sample like this?

## Move indices to another temperature node

If you want to move indices between nodes, you need to set additional settings.

1. the temperature information for the node must be set to `opensearch.yml`

```yml
node.attr.temp: hot
```

or

```yml
node.attr.temp: warm
```

2. Add `allocation.require.temp` as an action to the state

```yml
{
  "name": "warm",
  "actions": [
    {
      "allocation": {
        "require": {
          "temp": "warm"
        }
      }
    }
  ]
}
```

andrross:
> but I think this is a picture of a simple 'state' change. so there may be a little confusion.

Agreed that is confusing. I actually think the "add replicas after 1 day to increase read performance" is probably an uncommon use case. What do you think?

How about changing this example to be what might be the more common "warm" use case? I'm thinking the description changes to:

> In this case, an index is initially in a hot state. After a 7 days, it changes to a warm state, where the number of replicas is reduced to 1 and the indexes are moved to nodes with the "warm" attribute.
> 
> After 30 days, the policy moves this index into a delete state. The service sends a notification to a Chime room that the index is being deleted, and then permanently deletes it.

We'll need to change the policy details and diagram to match and also include the temperature details you have above. @10000-ki what do you think? Would you be willing to make those changes in this PR? I think @vagimeli can probably help with the diagram changes if needed.

10000-ki:
> Agreed that is confusing. I actually think the "add replicas after 1 day to increase read performance" is probably an uncommon use case. What do you think?

yes i agree

The warm area is less frequently read than the hot area, so rather than an example for raising read performance in the warm area.
It seems like a good idea to write a description of moving past indexes to the warm area.

> How about changing this example to be what might be the more common "warm" use case? I'm thinking the description changes to:

very nice 

> After a 7 days, it changes to a warm state

I think the above example is appropriate

> Would you be willing to make those changes in this PR

yes i can

10000-ki:
i will make a new pr

Issue #6339, opened 2024-02-02
[DOC] Availability and recovery Snapshots Take and restore snapshots 


etgraylog:
**What do you want to do?**
 
- [x] Request a change to existing documentation

**Tell us about your request.** Provide a summary of the request and all versions that are affected.
 
When an AWS EC2 Instance uses an AWS IAM Instance Profile with a Policy granting required access to the S3 Bucket and Trust Relationship established with ec2.amazonaws.com, adding `s3.client.default.access_key` & `s3.client.default.secret_key` to OpenSearch's keystore is not necessary & will cause problems if done while following the instructions.

Steps 2 thru 7 of these instructions can be skipped with this configuration:
https://opensearch.org/docs/2.11/tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore/#amazon-s3

Affected versions: 2.11.0 (& above)

**What other resources are available?** Provide links to related issues, POCs, steps for testing, etc.

hdhalter:
Thanks, @etgraylog ! Can you please submit a PR with the proposed changes?

PR #6349, opened 2024-02-05
Added 1x sentence to clarify use/support for IAM Instance Profiles.

etgraylog:
### Description
This adds a notation to the documentation for AWS IAM Instance profiles and their use with  OpenSearch Snapshots.

### Issues Resolved
Closes https://github.com/opensearch-project/documentation-website/issues/6339

### Checklist


_tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore.md
```diff
@@ -101,6 +101,8 @@ You will most likely not need to specify any parameters except for `location`. F
 
    After the Docker cluster starts, skip to step 7.
 
+   If you're using AWS IAM Instance Profiles to allow OpenSearch nodes on AWS EC2 instances to inherit Roles for Policies granting access to AWS S3 buckets, you can skip to step 8.
```
Naarcha-AWS, line 104:
```suggestion
   If you're using [AWS IAM instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) to allow OpenSearch nodes on AWS EC2 instances to inherit roles for policies when granting access to AWS S3 buckets, skip to step 8.

Naarcha-AWS (changes requested):

_tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore.md
Naarcha-AWS: @scrawfor99, can you verify that this is technically accurate?

scrawfor99: Hi @Naarcha-AWS, I don't have any true expertise in this area. However, judging from the existing documentation on the snapshots docs and the linked instance profile documentation this sounds reasonable. I am not able to confirm this without testing though so perhaps @etgraylog could share whether they have gone through these steps themselves (I would assume so) and that will work for verification.

etgraylog: > Hi @Naarcha-AWS, I don't have any true expertise in this area. However, judging from the existing documentation on the snapshots docs and the linked instance profile documentation this sounds reasonable. I am not able to confirm this without testing though so perhaps @etgraylog could share whether they have gone through these steps themselves (I would assume so) and that will work for verification.

Thank you @scrawfor99! Yes exactly, this discovery is from troubleshooting a deployment that makes use of an AWS IAM Instance Profile (that grants access to the S3 bucket of focus). When storing AWS IAM Role credentials (for S3 bucket access) in the OpenSearch Keystore, which the EC2 Instance(s) already had the ability to access via IAM Instance Profile granting the AWS IAM Role, failures would occur when attempting to access the bucket. After removing them from the OpenSearch Keystore while leaving the IAM Instance Profile, etc... in-place, access to the S3 Bucket worked as expected.

And thank you @Naarcha-AWS for adding the hyperlink - good idea! 👍 I'll sign off and commit that suggestion.

Naarcha-AWS (approved):

Issue #6363, opened 2024-02-06
[DOC] Views


peternied:
**What do you want to do?**
 
- [ ] Request a change to existing documentation
- [X] Add new documentation
- [ ] Report a technical problem with the documentation
- [ ] Other 

**Tell us about your request.** Provide a summary of the request and all versions that are affected.
 I'm adding a new set of administration and usage APIs for a new feature views.  This feature abstracts where data is from those that want to search it, like an alias with more flexibility.  The feature is experimental and will be building up a critical mass of functionality before becoming part of the public API.

**What other resources are available?** Provide links to related issues, POCs, steps for testing, etc.
- https://github.com/opensearch-project/OpenSearch/issues/6181

Note; I'll start writing documentation in draft for developers to review.

hdhalter:
Thanks, Peter. Do you have a release you are targeting?

peternied:
@hdhalter Not at this time, I'll update as soon as we've got it slated.

peternied:
I'm going to close out this issue until we've got investment in views - just like the PR that was closed

PR #6394, opened 2024-02-12
[Draft] Views documentation

peternied:
### Description
Adds documentation for the views feature

### Issues Resolved
- Resolves https://github.com/opensearch-project/documentation-website/issues/6363

### Checklist

### DIFF
--- a/_api-reference/index-apis/views.md
+++ b/_api-reference/index-apis/views.md
@@ -1,0 +1,180 @@
+---
+layout: default
+title: View APIs
+parent: Index APIs
+nav_order: 75
+redirect_from:
+  - /opensearch/rest-api/index-apis/views/
+---
+
+## Views
+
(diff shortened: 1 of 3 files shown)


peternied:
> @peternied maybe I missed it, but could you add a really clear explanation about the interaction between Views and roles? I think we want to make it really clear that we still use the normal permission system behind the scenes otherwise it may seem like Views can skip the normal check or are an alternative permission system.  

I have documented the feature as it is, as it changes I'll need to revise this PR. 

There is no change or loosening to any permissions - you'll need both index permissions as normal in addition to cluster permissions that allow calling the different view APIs

peternied:
I'm going to close this out - until there is a clear delivery data associated with this feature it doesn't make sense to clog up the PR queue.  I'll reopen if it becomes relevant in the future.

PR #7038, opened 2024-04-27
Correct docker-compose demo docs

reify-travis-brown:
Security is enabled by default so setting this environment variable has no effect unless the intent is to disable it for demo purposes.

### Description
This change allows for a demo to be started without additional security setup. Without this change, the current instructions lead to an invalid security state that cannot be used until further security configuration is completed.

### Issues Resolved
N/A

### Checklist


hdhalter:
Thanks, @reify-travis-brown ! 

@scrawfor99, @peterzhuamazon - Can you  approve for technical accuracy? Thanks!

scrawfor99 (changes requested):
Setting `DISABLE_SECURITY_PLUGIN` to `true` will cause the Security plugin to not be used. 

It is a little confusing because it is a double negative but I think we want to keep things as they are unless I am misunderstanding the intent of this change.

_security/configuration/demo-configuration.md
```diff
@@ -27,7 +27,7 @@ The demo configuration is automatically called as part of the setup for each sup
 Use the following steps to set up the Security plugin using Docker:
 
 1. Download [docker-compose.yml](https://opensearch.org/downloads.html).
-2. In the `docker-compose.yml` file, set `DISABLE_SECURITY_PLUGIN` to `false`.
+2. In the `docker-compose.yml` file, set environment variable `DISABLE_SECURITY_PLUGIN` to `true`.
```
peterzhuamazon, line 30: This section should be changed to `if user want to disable security plugin then set to true` or similar.

More details can be seen here:
https://github.com/opensearch-project/opensearch-build/tree/main/docker/release#disable-security-plugin-security-dashboards-plugin-security-demo-configurations-and-related-configurations

We have security always enable by default so no need to specifically make it false.
And this commit of making it to true will just disable the security by default, which is not the recommended way for docker.

peterzhuamazon (changes requested):

_security/configuration/demo-configuration.md
peterzhuamazon: The correct way to fix this is just changing `Use the following steps to set up the Security plugin using Docker` to something like `Use the following steps to disable Security plugin when using Docker`, or just directly link to the readme file for docker release in build repo:
https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/README.md#disable-security-plugin-security-dashboards-plugin-security-demo-configurations-and-related-configurations

We do not want to recommend users to disable security unless they specifically want to do so.

hdhalter: @peterzhuamazon - Let me know if this works. I took that line out of the process steps and moved it below.

_security/configuration/demo-configuration.md
```diff
@@ -27,14 +27,15 @@ The demo configuration is automatically called as part of the setup for each sup
...
-2. In the `docker-compose.yml` file, set `DISABLE_SECURITY_PLUGIN` to `false`.
-3. Run the following command:
+2. Run the following command:
 
 ```bash
 docker-compose up
 ```
 {% include copy.html %}
 
+If you want to disable the Security plugin when using Docker, set the `DISABLE_SECURITY_PLUGIN` environment variable  to `true` in the `docker-compose.yml` file. Disabling the Security plugin is not recommended. For more information, see the [Docker release notes](https://github.com/opensearch-project/opensearch-build/tree/main/docker/release#disable-security-plugin-security-dashboards-plugin-security-demo-configurations-and-related-configurations) in GitHub.
```
peterzhuamazon, line 37: Hi @reify-travis-brown could you rename `Docker release notes` to `Docker image release README`?

Thanks.

hdhalter, line 37:
```suggestion
If you want to disable the Security plugin when using Docker, set the `DISABLE_SECURITY_PLUGIN` environment variable  to `true` in the `docker-compose.yml` file. Disabling the Security plugin is not recommended. For more information, see the [Docker image release README](https://github.com/opensearch-project/opensearch-build/tree/main/docker/release#disable-security-plugin-security-dashboards-plugin-security-demo-configurations-and-related-configurations) in GitHub.
```

peterzhuamazon (approved):

scrawfor99 (approved):

PR #7039, opened 2024-04-28
Update the example of the multiple terms aggregation by muliple orders.

### LINKED ISSUE #7037: [DOC] Update the document for multiple terms aggregation
- [ ] Add new documentation
- [ ] Other 

The current document about the multiple terms aggregation is not accurate. In the document(https://opensearch.org/docs/latest/aggregations/bucket/multi-terms/), it gives an example. And it implies that multiple terms aggregation only accepts one order. But in fact, it accepts multiple orders like the following:
```
GET sample-index100/_search
{
  "size": 0, 
  "aggs": {
    "hot": {
      "multi_terms": {
        "terms": [{
          "field": "region" 
        },{
          "field": "host" 
        }],
        "order":[ {"max-cpu": "desc"},  {"max-memory": "desc"}] <=== multiple orders
      },
      "aggs": {
        "max-cpu": { "max": { "field": "cpu" } },
        "max-memory":  { "max": { "field": "memory" } }
      }      
    }
  }
}
```
Please help to update the document, and I think this is an important and valuable feature for the multiple terms aggregation.

Zhikai-VM:
### Description
Update the example of the multiple terms aggregation.

### Issues Resolved
Closes #7037

### Checklist


ankyhe:
I suggest we don't remove the original example, but add a new one.  Because if it needs only 1 order, it could use original one.  If we change document like this way, it could confuse the old order style is not supported.  But it is not true, it can support both one order with `{...}` or multiple order with `[...]`.

Zhikai-VM:
> I suggest we don't remove the original example, but add a new one. Because if it needs only 1 order, it could use original one. If we change document like this way, it could confuse the old order style is not supported. But it is not true, it can support both one order with `{...}` or multiple order with `[...]`.

I got your point, but the current document is not correct. I am trying to fix it to make it accurate.

hdhalter:
@bowenlan-amzn - What are your thoughts on this?

Zhikai-VM:
> #7037

Please refer to the update in PR: https://github.com/opensearch-project/OpenSearch/pull/13400

Zhikai-VM:
> @bowenlan-amzn - What are your thoughts on this?

Please refer to the update in PR: https://github.com/opensearch-project/OpenSearch/pull/13400

hdhalter:
Hi @dblock - Can you please comment with your thoughts? Thanks.

dblock:
The current example is incorrect, if I understand correctly it has to be an array of orders, even with 1 order. There's no point in having an array of 1 and then another example with an array of 2, is there?

Zhikai-VM:
> The current example is incorrect, if I understand correctly it has to be an array of orders, even with 1 order. There's no point in having an array of 1 and then another example with an array of 2, is there?

As far as I know, there is no example of array in the offical doc.

ankyhe:
> The current example is incorrect, if I understand correctly it has to be an array of orders, even with 1 order. There's no point in having an array of 1 and then another example with an array of 2, is there?

@dblock As current official document, the multi-term example is as below:
```
GET sample-index100/_search
{
  "size": 0, 
  "aggs": {
    "hot": {
      "multi_terms": {
        "terms": [{
          "field": "region" 
        },{
          "field": "host" 
        }],
        "order": {"max-cpu": "desc"}
      },
      "aggs": {
        "max-cpu": { "max": { "field": "cpu" } }
      }      
    }
  }
}
```
@Zhikai-VM wants to revise it as:
```
GET sample-index100/_search
{
  "size": 0, 
  "aggs": {
    "hot": {
      "multi_terms": {
        "terms": [{
          "field": "region" 
        },{
          "field": "host" 
        }],
        "order": [ {"max-cpu": "desc"}, {"max-mem": "asc"}]
      },
      "aggs": {
        "max-cpu": { "max": { "field": "cpu" } },
        "max-mem": { "min": { "field": "mem"}}
      }      
    }
  }
}
```
to keep the official document accordance with this PR: https://github.com/opensearch-project/OpenSearch/pull/13400

Zhikai-VM:
Hi @dblock , what do you think about this? I prefer to use an array for the 'order' field. We can pass one order or multiple orders in this field.

dblock:
This looks good to me, can we merge @hdhalter ?

Naarcha-AWS (approved):

Zhikai-VM:
Thanks @dblock @bowenlan-amzn!

PR #7144, opened 2024-05-14
[W.I.P.] Ubi integration

RasonJ:
### Description
Integrates User Behavior Insights logging in the documentation website.

### Issues Resolved
Tracks individual user interactions with the documentation website, so that we can use analytics to optimize the site utility for other users.

Events tracked:
- all gtags are copied into ubi events index
- search terms and results
- item_clicked
- page_exit
- dwell time
- 404 redirect

### Checklist
- [ x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and subject to the [Developers Certificate of Origin](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


epugh:
@RasonJ if this is a WIP, you can click the "Convert to Draft" and then remove the `[W.I.P]` label...

assets/js/search.js
```diff
@@ -206,6 +222,11 @@
             const searchResultClassName = 'top-banner-search--field-with-results--field--wrapper--search-component--search-results--result';
             if (!node || !_showingResults || node.classList.contains(CLASSNAME_HIGHLIGHTED)) return;
 
+            const link = node.querySelector('a');
```
epugh, line 225: I think we should either move forward on this, or remove it and add it in a follow up PR....

assets/js/search.js
```diff
@@ -247,14 +268,103 @@
...
+
         const navToHighlightedResult = () => {
             const searchResultClassName = 'top-banner-search--field-with-results--field--wrapper--search-component--search-results--result';
             elResults.querySelector(`.${searchResultClassName}.highlighted a[href]`)?.click?.();
         };
 
+        /**
+         * Find item and position clicked
+         * Modifies the ubi event data if the item is found
+         * @param {*} ubiEvent
```
epugh, line 291: A bit more detail on the `@param` since we have it?

assets/js/search.js
```diff
@@ -247,14 +268,103 @@
...
+                        }
+                    }
+                }
+            }
+            return ubiEvent;
+        };
+
+
+        const makeUbiEvent = (name, event_type, data) => {
+            let e = new UBI.UbiEvent(name);
```
epugh, line 330: I think either we add docs to all these, or have no docs.  If we have some short docs, it might help "the next person" learn how to use UBI.

assets/js/search.js
```diff
@@ -247,14 +268,103 @@
...
+            }
+
+            if(name == 'search'){
+                e.message_type = 'QUERY';
+                e.message = data.search_term;
+                e.event_attributes.object = data;
+            } else if(name == 'item_click') {
+                e = setUbiClickData(e, data);
+            } else {
+                switch(event_type) {
```
epugh, line 343: Likewise I'd like to see this switch statement filled in or removed....   To get this to something that is committable.

assets/js/ubi.js
```diff
@@ -0,0 +1,302 @@
+
+
+
+
+function guiid() {
```
epugh, line 5: This method always struck me as an odd name.   It's generading a GUID right?    `guid`?

assets/js/ubi.js
```diff
@@ -0,0 +1,302 @@
...
+		event_attributes:{dwell_seconds:seconds},
+		data_object:TimeMe
+	});
+	logEvent(e);
+}
+
+export async function logEvent(event){
+	try {
+		//=>146.190.147.150
+		fetch('http://localhost:9200/ubi/docs', {
```
epugh, line 202: We do need to figure out how we handle this!

epugh (changes requested):
I'm going to give this a spin locally, and I'm going to take a stab at the whitespace formatting ;-). My eyes hurt with the combination of 2, 4 and 8 spacing.

_layouts/default.html
```diff
@@ -254,8 +254,26 @@ <h2>Related articles</h2>
       anchors.add().remove('.subfooter h1, .subfooter h2');
     </script>
   {% endif %}
+  <script src="{{ '/assets/js/timeme.min.js' | relative_url }}"></script>
+  <script type="text/javascript">
+    // Initialize library and start tracking time
```
epugh, line 259: @RasonJ why do we have this initializing twice?   Also, I don't think we need the comments...  Normally I love comments, but what it does appears pretty clear...

epugh (changes requested):

RasonJ:
The default and home html views are not both on each full webpage.  So, it makes sure that it’s initialized on most pages.

assets/js/ubi.js
```diff
@@ -0,0 +1,300 @@
...
+  h1 ^= Math.imul(h2 ^ (h2 >>> 13), 3266489909);
+  h2  = Math.imul(h2 ^ (h2 >>> 16), 2246822507);
+  h2 ^= Math.imul(h1 ^ (h1 >>> 13), 3266489909);
+
+  return 4294967296 * (2097151 & h2) + (h1 >>> 0);
+};
+
+
+/**
+ * In place of true authentication, this makes a hash out of the user's IP address and browser
```
AMoo-Miki, line 31: Employees of large organizations usually use versions of the browsers imposed on to them and all of their connections would be from the IP of the organization. Does this mean they would all be recognized as the same user?

assets/js/ubi.js
```diff
@@ -0,0 +1,300 @@
...
+    rq.onerror = function(){
+      userError();
+      if(this.error != null && this.error != ''){
+        console.error('ERROR Retrieving user info: ' + this.error);
+      }
+      else
+        console.error('UNSPECIFIED ERROR Retrieving user info');
+    }
+
+    rq.open("GET", "https://api64.ipify.org?format=json", true);
```
AMoo-Miki, line 83: We shouldn't be making calls to external providers. It would be great if we could drop the whole IP-based approach.

assets/js/ubi.js
```diff
@@ -0,0 +1,300 @@
...
+    event_attributes:{dwell_seconds:seconds},
+    data_object:TimeMe
+  });
+  logEvent(e);
+}
+
+export async function logEvent(event){
+  try {
+    //=>146.190.147.150
+    fetch('http://localhost:9200/ubi/docs', {
```
AMoo-Miki, line 200: This comment is to track WiP item.

assets/js/ubi.js
```diff
@@ -0,0 +1,300 @@
...
+      method: 'POST',
+      headers: {
+      'Accept': 'application/json, text/plain, */*',
+      'Content-Type': 'application/json'
+      },
+      body: event.toJson()
+    }).then(res => res.json())
+    .then(res => console.log(res))
```
AMoo-Miki, line 208: This comment is to track WiP item.

assets/js/ubi.js
RasonJ: Yep.  It was the best we could think of without having any access to the webserver code.  Part of User Behavior Insights is tracking individual users (and their behaviors).  This method is obviously not guaranteed to be unique, but it's unique-ish to the ip, and should provide some individual signal.  We figured using the google api was reasonable since you're already using google analytics.
However, since you are far more familiar with the production environment, if you can think of other ways to track individual users, we'll be happy to include that instead.

assets/js/ubi.js
```diff
@@ -164,23 +147,48 @@ export function getPageId(){
...
+      trail += ` › ${window.location.pathname}`;
+    }
+  } else {
+    trail = window.location.pathname;
+  }
+  sessionStorage.setItem('trail', trail);
+
+  return trail;
+}
+
```
AMoo-Miki, line 170: We would be better off with a reliable structure. Also, `startsWith` logic doesn't sound right. How about:
```suggestion
function getTrail(){
  try {
    const trail = JSON.parse(sessionStorage.getItem('trail'));
    if (Array.isArray(trail)) return trail;
  } catch (ex) { /* Do nothing */ }

  return [];
}

function setTrail(){
  const trail = getTrail();
  // No need to add the current pathname if it is already the last element in trail
  if (trail.length && trail[trail.length - 1] === window.location.pathname)
    return trail;

  sessionStorage.setItem('trail', trail.concat(window.location.pathname));

  return trail;
}

```

assets/js/search.js
RasonJ: What do you mean?

assets/js/search.js
RasonJ: Do we want to log item hovers?

epugh: Let's skip for now, for that exact data reason.

RasonJ:
Test implementation here: http://137.184.176.129:4000/docs/latest/
Dashboards for the implementation are also on that server: http://137.184.176.129:5601/app/opensearch-query-workbench#/

hdhalter:
@epugh , @RasonJ - Do we still need this PR for 2.15?

RasonJ:
> @epugh , @RasonJ - Do we still need this PR for 2.15?
I'm not sure.  @epugh, any contractual/financial  reason why we would need this for 2.15? 
This is probably the least urgent of our PR's, but the sooner there is more realistic data from the site, the sooner we could potentially improve/refactor UBI.  Also, at some point I suspect OSC will stop hosting the sandbox server.

epugh:
This PR can stay draft...   It becomes important once we have the OpenSearch cluster that powers the documentation website migrated to managed services version 2.15 or later.   That isn't planned till September (IIRC), and then this becomes something mergable.   In the meantime, this is a nice example of using UBI, and we will keep it up to date and deployed on the demo server.

kolchfa-aws:
@RasonJ Closing this as stale for now. Please reopen when it's ready for review. Thanks!

PR #7538, opened 2024-06-26
Drop repeated key/word in inverted index

xofbd:
"the" appears twice as a key in the inverted index; keys should only appear once. The correct key-value pair is kept, with a value of "1, 2" as "the" appears in both documents.

### Description
It looks like "the" is repeated twice in the inverted index. From my understanding, all keys must be unique. The correct key-value pair should be `"the": 1, 2` as "the" appears in both documents.

### Issues Resolved
No issue was created.

### Version
No version is necessary since this is a minor typo correction.

### Frontend features
Not applicable

### Checklist

### DIFF
--- a/_getting-started/intro.md
+++ b/_getting-started/intro.md
@@ -106,7 +106,6 @@ in | 1
 the | 1, 2
 eye | 1
 of | 1
-the | 1
 beholder | 1
 and | 2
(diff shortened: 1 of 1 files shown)


kolchfa-aws (approved):
Thank you, @xofbd!

PR #8363, opened 2024-09-24
Disambiguate statement "value defaults to a fixed percentage"

drewmiranda-gl:
This page documents that "In the case of a dedicated search node where the node exclusively has the search role, this value defaults to a fixed percentage of available storage."

However, the document does not provide specifics about what this fixed percentage is. From what I can surmise reading the code [[1](https://github.com/opensearch-project/OpenSearch/blob/b984b9f56f9b0f9dbb5322bddf04f3b45940b730/server/src/main/java/org/opensearch/node/Node.java#L396)] [[2](https://github.com/opensearch-project/OpenSearch/blob/b984b9f56f9b0f9dbb5322bddf04f3b45940b730/server/src/main/java/org/opensearch/node/Node.java#L2105)], and from doing real world testing, this value is set to 80% of the available space on the volume where the file cache lives.

### Description

Clarify specifically what fixed percentage is used. See above.

### Version

All.

### Checklist


Naarcha-AWS (approved):
Verified and approved.

PR #9063, opened 2025-01-14
Add guidance for using transformation during migration

peternied:
### Description
Add guidance for migrations that need to use the transformation systems.

### Version
The most recent version of the Migration Assistant.

### Checklist

### DIFF
--- a/_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
+++ b/_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
@@ -20,7 +20,7 @@ During a migration there is an opportunity change names and paths of index and d
 #### Migration with a snapshot
 
 1. Navigate to the bootstrap box and open the `cdk.context.json` with vim
-2. Add/Update the key `reindexFromSnapshotExtraArgs` to include `--transformer-config-file /shared-logs-output/rfs-transform.json`
+2. Add/Update the key `reindexFromSnapshotExtraArgs` to include `--doc-transformer-config-file /shared-logs-output/rfs-transform.json`
 3. Redeploy the Migration Assistant 
 4. Navigate on to the Migration Assistant console
 5. Create a file with `vim /shared-logs-output/rfs-transform.json`
@@ -31,7 +31,7 @@ During a migration there is an opportunity change names and paths of index and d
     "JsonConditionalTransformerProvider": [
       {
         "JsonJMESPathPredicateProvider": {
-          "script": "index._index == '{{INDEX_ORIGINAL_NAME}}"
+          "script": "index._index == '{{INDEX_ORIGINAL_NAME}}'"
         }
       },
       [
@@ -41,7 +41,7 @@ During a migration there is an opportunity change names and paths of index and d
               "operation": "modify-overwrite-beta",
               "spec": {
                 "index": {
-                  "\\_index": "INDEX_NEW_NAME"
+                  "\\_index": "{{INDEX_NEW_NAME}}"
                 }
               }
             }
@@ -81,5 +81,5 @@ During a migration there is an opportunity change names and paths of index and d
 ]
 ```
 10. Replace both `{{INDEX_ORIGINAL_NAME}}` and `{{INDEX_NEW_NAME}}`
-11. Run metadata migration with the additional parameter `console metadata migrate --transformer-config-file /shared-logs-output/metadata-transform.json`
+11. Run metadata migration with the additional parameter `console metadata migrate --doc-transformer-config-file /shared-logs-output/rfs-transform.json`
 12. Run backfill as normal
--- a/_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
+++ b/_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
@@ -10,7 +10,7 @@ redirect_from:
 
 # Adjusting settings during migration
 
-During a migration there is an opportunity change names and paths of index and data structures that are not modifiable after they are declared.  This page provides a cookbook of different scenarios and templates that can be used to make these adjustments.
+During a migration there is an opportunity change names and paths of index and data structures that are not modifiable after they are declared. This page provides a cookbook of different scenarios and templates that can be used to make these adjustments.
 
 ### Rename an index
 
--- a/_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
+++ b/_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
@@ -1,0 +1,85 @@
+---
+layout: default
+title: Adjusting settings during migration
+nav_order: 60
+parent: Planning your migration
+grand_parent: Migration phases
+redirect_from:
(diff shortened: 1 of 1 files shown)


_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
```diff
@@ -0,0 +1,85 @@
...
+grand_parent: Migration phases
+redirect_from:
+  - /migration-assistant/migration-phases/adjusting-settings-during-migration/
+---
+
+# Adjusting settings during migration
+
+During a migration there is an opportunity change names and paths of index and data structures that are not modifiable after they are declared. This page provides a cookbook of different scenarios and templates that can be used to make these adjustments.
+
+### Rename an index
```
gregschohn, line 15: Do we suspect that this will be a common requirement/ask from customers?  I suspect that most customers will either NOT want to rename any of their indices, or if they do, will have very specific reasons and requirements around doing that.

Is this guidance general enough that more than one customer (ideally, more than 10%) would benefit from it?  If not, I'd rather keep this in a set of internal runbooks for more targeted communications with users.

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
```diff
@@ -0,0 +1,85 @@
...
+1. Navigate to the bootstrap box and open the `cdk.context.json` with vim
+2. Add/Update the key `reindexFromSnapshotExtraArgs` to include `--doc-transformer-config-file /shared-logs-output/rfs-transform.json`
+3. Redeploy the Migration Assistant 
+4. Navigate on to the Migration Assistant console
+5. Create a file with `vim /shared-logs-output/rfs-transform.json`
+6. Add the following to the file:
+```json
+[
+  {
+    "JsonConditionalTransformerProvider": [
```
gregschohn, line 31: @AndreKurait - do you think we'll keep conditional transformers in place once we support writing transformations in scripting languages (e.g. javascript)?

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
```diff
@@ -0,0 +1,85 @@
...
+> [!NOTE]  
+> We recommend creating an alias to an index instead of renaming them, if an alias is created on the source cluster the migration process will copy the alias over as well. See create an alias [documentation](https://opensearch.org/docs/latest/api-reference/index-apis/update-alias/).
+
+#### Migration with a snapshot
+
+1. Navigate to the bootstrap box and open the `cdk.context.json` with vim
+2. Add/Update the key `reindexFromSnapshotExtraArgs` to include `--doc-transformer-config-file /shared-logs-output/rfs-transform.json`
+3. Redeploy the Migration Assistant 
+4. Navigate on to the Migration Assistant console
+5. Create a file with `vim /shared-logs-output/rfs-transform.json`
```
gregschohn, line 26: We shouldn't recommend to users that they WRITE a transform in a shared log.  It's arguable if that volume should be writable by the migration console.  Same argument for the other file below.

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
peternied: I think rename is common ask and its a simple and useful example case.  The only request that is more popular is how to change the shard count - requiring more thoughtful messaging of the calculations that I would prefer not to start with.

Do you think we should have a different scenario we could do that instead of this one?

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
peternied: This code is functional - when JavaScript becomes available we can always edit this example to be cleaner.

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
peternied: Thanks - what is an alternative way to accomplish this?

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
```diff
@@ -0,0 +1,85 @@
...
+          }
+        }
+      ]
+    ]
+  }
+]
+```
+10. Replace both `{{INDEX_ORIGINAL_NAME}}` and `{{INDEX_NEW_NAME}}`
+11. Run metadata migration with the additional parameter `console metadata migrate --doc-transformer-config-file /shared-logs-output/rfs-transform.json`
+12. Run backfill as normal
```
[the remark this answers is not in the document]
peternied: I don't know how to write one, would you like to add one so we can have both at the same time?

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
gregschohn: We've got approval on the UPL license yesterday, so I'd expect that Javascript will be available imminently (and jolt transforms will be removed probably at the 2.2 release).  Please replace these instructions with javascript ones so that this description isn't DOA and confusing to customers.

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
gregschohn: If you have a file, here's how you can convert it into json and then base64
```
jq -n --rawfile script1 ~/Downloads/es-load-test.py '{"fileContent": $script1}'| base64
```
That could then be pushed into the extra args for the applications ("--transformer-config-base64 ...").  Notice that we'll need the json to match what the JsonJSTransformerProvider expects in the PR ([see here](https://github.com/opensearch-project/opensearch-migrations/pull/1207/files#diff-2223bbb5e0fd42b5923d57314dc4741f98d8ed14136b23517995cc3d93cea243)).

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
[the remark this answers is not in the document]
gregschohn: If this is one of the most popular transforms (ahead of even migrating between legacy ES versions and new versions of OpenSearch), we should have a direct transformation to provide this functionality.  We should start with the interface for the TypeMappingsSanitization transform.  We should refactor some of the common logic between renaming indices w/ and without types together.  Have you created a jira for that?

peternied:
Thanks for the reviews, after talking this over we don't like the user experience and are going to come back with documentation after we've cleaned up the experience in Migration Assistant.

_migration-assistant/migration-phases/planning-your-migration/adjusting-settings-during-migration.md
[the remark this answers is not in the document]
gregschohn: added some AC to https://opensearch.atlassian.net/browse/MIGRATIONS-2356

PR #9381, opened 2025-03-10
Add clarity to request fields in reindex.md

cwadrupldijjit:
I noticed that the tabular format for the request fields made it confusing when reading the documentation.  From what it looks like, several fields needed to show that they are inside of the object(s) rather than directly in the request object.

This is only one possible solution to the problem, but I could also see it as confusing since things like index settings use dot-separated object keys (though they also seem to support actual objects rather than just the dot-separated key notation for the most part) and I'm not certain if it will accept that same flat-object, dot-separated key notation for these requests, either.

### Description
Fixes confusing documentation structure

### Issues Resolved
N/A

### Version
2.19 (at least)

### Frontend features
N/A

### Checklist


Naarcha-AWS (approved):

Naarcha-AWS:
@cwadrupldijjit: This will likely be addressed once we've merged more API automation based on the Spec. Going to keep this open for now and might loop back around to add the automation markers once we merge in #9385 is merged.

Naarcha-AWS:
Going to close this PR for now since this will be addressed while closing other content gaps.

PR #9667, opened 2025-04-21
Adds doc for Resource Sharing and Access Control feature

### LINKED ISSUE #9652: [DOC] Centralized Resource Access Control Framework
**What do you want to do?**
 
- [x] Add new documentation
- [ ] Other 

Security plugin introduces a new mechanism to offload Resource authz mechanism from plugins to security by exposing APIs that plugins can call. This feature also enables users to manage access to their own resources.

**Version:** 
3.0.0 and above
 
https://github.com/DarshitChanpura/security/blob/sample-resource-plugin-spi/RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md

DarshitChanpura:
### Description
Adds doc for the newly added Centralized Resource Access Control and Sharing feature.

### Issues Resolved
Closes #9652 

### Version
3.0

### Checklist


DarshitChanpura:
Local render: 
https://github.com/user-attachments/assets/04ddad65-8efb-4ae0-826c-84df96c4b246

DarshitChanpura:
Screenshot 2025-04-21 at 2 25 50 PM

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+
+The Resource Sharing extension in the OpenSearch Security plugin provides fine-grained, resource-level access management for plugin-declared resources. It builds on top of existing index-level authorization to enable you to:
+
+- Share and revoke access as a resource owner.
+- View and manage all resources as a super-admin.
+- Define custom shareable resources through a standardized Service Provider Interface (SPI).
+
+## Enabling the resource sharing extension
+
+To enable the resource sharing extension, follow these steps.
```
Naarcha-AWS, line 24:
```suggestion
To enable the resource sharing extension, set up the plugin and permission settings.
```

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+      "backend_roles": ["data_analyst"]
+    }
+  }
+}
+```
+{% include copy.html %}
+
+## Action groups
+
+Action groups define permission levels for shared resources, for example, `default`. To share resources across the action groups, use the `share_with` array in that resource's configuration and add wildcards for each default role:
```
Naarcha-AWS, line 134:
```suggestion
Action groups define permission levels for shared resources. As of OpenSearch 3.0, the only supposted action group for resource sharing is `default`. To share resources across the action groups, use the `share_with` array in that resource's configuration and add wildcards for each default role:
```

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+client.revoke(resourceId, indexName, recipients, listener);
+
+// List accessible IDs
+client.getAccessibleResourceIds(indexName, listener);
+```
+{% include copy.html %}
+
+### Document structure
+
+The resource sharing metadata is stored as JSON documents that define ownership, permissions, and access patterns. Each document contains fields that specify the resource location, identifier, creator information, and sharing configuration:
```
Naarcha-AWS, line 112: Where would I find or add this document? Is it an API call to the security plugin?

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+- Resources must reside in system indexes.
+- Disabling system index protection exposes resources to direct index-level access.
+
+## Best practices
+
+To ensure secure and efficient implementation of resource sharing and access control, follow these recommended practices:
+
+- Declare and register extensions correctly.
+- Use the SPI client APIs instead of manual index queries.
+- Enable only on fresh 3.0.0+ clusters to avoid upgrade issues.
```
Naarcha-AWS, line 170: What do we mean by "fresh" clusters?

Naarcha-AWS (changes requested):
@DarshitChanpura: Most of this content seemed like implementation notes, so I tried to remove anything related to the development of the extensions or the plugin and make it more user-specific. That being said, is this extension meant only for plugin developers? If so, that would change how where we place the content.

I also have a few technical questions that need clarification.

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+```
+{% include copy.html %}
+
+## Resource sharing components
+
+The resource sharing extension consists of key components that work together to provide standardized access management. The primary component is the Service Provider Interface (SPI), which serves as the foundation for plugin integration and resource management.
+
+### opensearch-security-spi
+
+The `opensearch-security-spi` component:
```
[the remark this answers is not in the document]
DarshitChanpura: Yes.

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Defines `ResourceSharingExtension` for plugin implementations.
+- Tracks registered resource plugins at startup.
+- Exposes a `ResourceSharingClient` for performing share, revoke, verify, and list operations.
+
+You can customize this component based on the SPI implementation. To customize, create an extension file `src/main/resources/META-INF/services/org.opensearch.security.spi.ResourceSharingExtension` containing your SPI's class name:
```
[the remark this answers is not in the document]
DarshitChanpura: This is purely for plugin developers. It will be used when a new/existing plugin declares themselves as Resource plugin. Similar model is already in place for JobScheduler.

_security/access-control/resources.md
DarshitChanpura: This is not visible to regular users directly as it is hidden behind SystemIndex Protection.

For plugins intending to declare a protected resource, they can retrieve the information about who can access this resources by simply calling security client's hasPermission.

End users will, in future, iterations be able to control who they can share the resource with once plugin implements those APIs and makes the info available via dashboard.

At this point, the documentation intends to serve plugin developers, with allowing admins to optionally opt in to this feature by enabling feature flag.

We are trying to implement a sharing model that doesn't exist in OpenSearch yet. This is the first step by centralizing the sharing information to security, rather than plugins storing it individually.

_security/access-control/resources.md
DarshitChanpura: This feature is intended to be enabled only on brand new clusters.

DarshitChanpura: In next iteration, we will add a migration path to enable usage on existing clusters.

DarshitChanpura:
Yes, the extension is meant for plugin developers only. The feature itself toggles the "centralized " vs "non-centralized " approach when it comes to the internal sharing model.

For example, for new cluster, if AD chooses to use this and their filter by backend role setting, all detectors will upon creation be shared with the creators backend roles. However, the difference between old way and the new way is that with new way the sharing information will be stored in central index in security vs with old approach it will be stored in the ad-config index.

_security/access-control/resources.md
Naarcha-AWS: What would the `hasPermission` call look like?

DarshitChanpura: Plugins would call the client by passing the id of the resources, the index name where the resource sits and a listener to receive the response. This will then be evaluated on security plugin side and a boolean response will be sent through the listener.

_security/access-control/resources.md
```diff
@@ -0,0 +1,173 @@
...
+```yaml
+sample_full_access:
+  cluster_permissions:
+    - 'cluster:admin/sample-resource-plugin/*'
+
+sample_read_access:
+  cluster_permissions:
+    - 'cluster:admin/sample-resource-plugin/get'
+```
+{% include copy.html %}
```
nibix, line 57: To which users shall these roles be assigned to? Anyone who shall be able to do resource sharing?

nibix:
Thank you for this PR!

Like @Naarcha-AWS I also believe that is mostly developer docs; I think we should try to keep user documentation and developer documentation separate. 

User documentation for this feature can be very minimalisting IMHO:

- Notice about experimental status
- How to enable
- Situation with security disabled/enabled/not present
- Necessary permissions
- How to use 
  -  IMHO, this will be mostly just referring readers to the docs of the respective plugins utilizing the features.
  - Do we have any REST APIs left that are exposed by the security plugin regarding resource sharing? If so, these should be documented.
- Did I forget something?

Developer documentation:

A separate documentation target at developers should contain:

- Everything that touches Java classes, interfaces, SPIs. 

I am not exactly sure where such docs should go to. There is https://docs.opensearch.org/docs/latest/developer-documentation/ but this mostly points to .md files in the git repos. There are just two small documents that are parts of the documentation website. What's the general strategy regarding that? Shall the section on the documentation website be extended? My personal opinion would be that it is better to keep it directly in respective project repos, as this minimizes the gap between code and docs. But that's just IMHO.

Naarcha-AWS:
@nibix: Traditionally, Developer Documentation would belong in the repository where the source code lives, in this case, OpenSearch Security? If that's the correct location @DarshitChanpura, I can try and open a PR against that repo where the classes and such should live.

_security/access-control/resources.md
```diff
@@ -0,0 +1,139 @@
...
+{: .tip }
+
+### Cluster permissions
+
+The Security plugin must have the following access:
+
+- Permissions to make share, verify, and list requests.
+- Shared access to all cluster components, or be the owner of the cluster.
+
+To grant the resource sharing extension these permissions, add the following role to `roles.yaml`:
```
Naarcha-AWS, line 46:
```suggestion
To grant the resource sharing extension these permissions, add the following roles to `roles.yaml`. These roles will be assigned to anyone conducting resource sharing:
```

_security/access-control/resources.md
```diff
@@ -0,0 +1,139 @@
...
+      "backend_roles": ["data_analyst"]
+    }
+  }
+}
+```
+{% include copy.html %}
+
+## Action groups
+
+Action groups define permission levels for shared resources. As of OpenSearch 3.0, the only supposted action group for resource sharing is `default`. To share resources across the action groups, use the `share_with` array in that resource's configuration and add wildcards for each default role:
```
Naarcha-AWS, line 100:
```suggestion
Action groups define permission levels for shared resources. As of OpenSearch 3.0, the only supported action group for resource sharing is `default`. To share resources across the action groups, use the `share_with` array in that resource's configuration and add wildcards for each default role:
```

Naarcha-AWS (approved):

Naarcha-AWS:
Moving this out of 3.0.0

kolchfa-aws:
@DarshitChanpura This definitely belongs in developer documentation for the repo that the code is implemented in. If you'd like, we can list it in the developer documentation section of our documentation, which is pretty rudimentary. https://docs.opensearch.org/docs/latest/developer-documentation/. Closing this PR for now.

PR #9110, opened 2025-01-24
Add markers for cluster APIs.

Naarcha-AWS:
Part of issue #9050 

### Checklist
- [x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and subject to the [Developers Certificate of Origin](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #9811, opened 2025-05-04
Extend _cluster docs for _cluster/reroute and _cluster/state

Aeternax:
### Description
Added docs for _cluster/state and _cluster/reroute.
In particular _cluster/reroute is useful for people once they run into failed allocations.
(Which was how I found out that it's missing in the docs)

### Issues Resolved
Does not close a whole issue, but part of #3614


Naarcha-AWS:
@Aeternax: We already have a PR open for these APIs which includes automation from the API spec.

#9110

PR #10545, opened 2025-07-31
correct ISM job_interval  description and default

Aeternax:
### Description
This corrects the table on job_intervals, as it lists "5 minutes" as a value, which doesn't work.
Not much else to say here. 

### Version
all (none, just documentation)


Aeternax:
> Thanks, @Aeternax! Do other time values (like `plugins.index_state_management.coordinator.sweep_period` and `plugins.index_state_management.coordinator.backoff_millis`) need to be modified as well? Also, could you please fix DCO so we can merge the PR?

Hi @kolchfa-aws, I honestly didn't really check/think about it before, just ran into this setting.
But I just updated some other values to the correct defaults and formats as you pondered.

kolchfa-aws (approved):
LGTM. Thank you, @Aeternax!

PR #11638, opened 2025-12-11
Add dark mode to the documentation website

kolchfa-aws:
Addressing multiple requests, adding dark mode to the doc site.

dark-mode

### Checklist

### DIFF
--- a/_sass/_navigation-header.scss
+++ b/_sass/_navigation-header.scss
@@ -369,7 +369,7 @@ $max-container: 1440px;
                                 }
                                 @media screen and (max-width: $header-banner-mobile-breakpoint) {
                                     @include body-text;
-                                    color: #FFFFFF;
+                                    color: var(--nav-link-color);
                                     padding-left: 30px;
                                     &.in-category {
                                         color: var(--nav-link-hover-color);


ashwin-pc:
This does work, but given that this PR is meant to only enable dark mode, a lot of the changes here are actually not necessary and not the best way to do it. 

E.g. you make a lot of changes to the colors but changing them from using a generic value like $primary-xxx to a specific color value like $blue-300 and then change what blue -300 means in dark mode.

I don't know too much of this repo, but usually when we do dark/light mode, we keep the color variables constant since they have a specific meaning and update $primary-xxx to use either $blue-300 or $blue-100 for light and dark mode. That way we never go off brand by using color values that aren't a part of the brand and make sure all the primary color ($blue-100 etc) values are part of the brand colors

There are also a lot of other changes that don't seem to be related to dark mode. Also not an issue if they are intended, but it's better to make sure that all the changes in a PR are related to the feature being reviewed and nothing else.

kolchfa-aws:
@ashwin-pc Thank you for reviewing! This PR also includes a refactoring of existing code. The custom.scss file had a lot of duplication of the same styles that I extracted into variables. I also extracted button styling so all buttons can use the same style. The custom.scss was hard to follow and extend in its previous state.

kolchfa-aws:
@ashwin-pc I addressed your comments - the variables are now semantically named and have different values depending on the mode.

_sass/_navigation-header.scss
```diff
@@ -358,31 +360,31 @@ $max-container: 1440px;
...
-                                        text-decoration-color: $primary-open-sky-s2;
+                                        text-decoration-color: var(--nav-link-hover-color);
                                         text-decoration-thickness: 2px;
                                         text-underline-offset: 6px;
                                     }
                                 }
                                 @media screen and (max-width: $header-banner-mobile-breakpoint) {
                                     @include body-text;
-                                    color: $primary-open-sky-t3;
+                                    color: #FFFFFF;
```
ashwin-pc: This will be an issue in light mode right?

assets/js/_version-selector.js
```diff
@@ -138,7 +138,7 @@ const tpl = `
         padding: 0.3em calc(3.8em - 1px) 0.3em calc(1em - 1px);
         border-bottom: 1px solid #eee;
         text-decoration: none;
-        color: var(--link-color);
+        color: #0053b8;
```
ashwin-pc, line 141: we dont need this color to be different between light and dark mode?

_sass/_footer.scss
```diff
@@ -105,8 +108,8 @@ $footer-1-column-max:700px;
 
     h2 {
         font-size: 16px;
-        border-top: 1px solid $line;
-        color: $background-lightest;
+        border-top: 1px solid $grey-lt-300;
+        color: $grey-lt-000;
```
ashwin-pc, line 112: Are these strict colors here intentional. Im highlighting just this but i see another one on line 13 which says `$blue-300` and a few others which based on my understading will be the same color in light and dark mode. Is that expected? basically just making sure that anywhere you reference the color value directly, its intentional. 

Even the gradient start and end also use statuc colors irrespective of the theme

ashwin-pc (approved):
Looks good to me. Just confirm that the variables used in some places that have static values is intentional.

_sass/_navigation-header.scss
kolchfa-aws: This is the default color that's overridden later but I've changed it to use the variable - thanks.

assets/js/_version-selector.js
kolchfa-aws: The version selector stays the same color in both themes so this doesn't need to change.

_sass/_footer.scss
kolchfa-aws: The footer and home banner stay the same for both themes so they don't need to change colors. All other gradients use the variables.

Issue #12079, opened 2026-03-10
[DOC] Document hybrid query behavior when nested inside function_score


martin-gaievski:
**What do you want to do?**
 
- [X] Request a change to existing documentation
- [ ] Add new documentation
- [ ] Report a technical problem with the documentation
- [ ] Other 

**Tell us about your request.** 
Hybrid query documentation should document the behavior when hybrid query is used inside a `function_score` query.

Starting from OpenSearch 3.6 (PR opensearch-project/neural-search#1791), hybrid query can be nested inside `function_score`. However, the search pipeline's normalization/combination processor is not applied in this configuration. The sub-query scores are combined as a raw boolean `should` sum, and then the `function_score` functions (field_value_factor, script_score, etc.) are applied on top.

In the [hybrid query documentation](https://docs.opensearch.org/latest/query-dsl/compound/hybrid/):

1. Add a section or note explaining that `function_score` wrapping is supported but normalization is bypassed
2. Example query showing `function_score` with nested hybrid query
3. Clarify that combination weights specified in the normalization processor are not applied in this configuration
4. Recommend using `function_score` on individual sub-queries or applying score boosting at the application level if normalized scores with function boosting are needed

Example query

```json
GET /my-index/_search?search_pipeline=my-pipeline
{
  "query": {
    "function_score": {
      "functions": [{
        "field_value_factor": {
          "field": "popularity",
          "modifier": "log2p"
        }
      }],
      "query": {
        "hybrid": {
          "queries": [
            {"match": {"title": "search terms"}},
            {"neural": {"embedding": {"query_text": "search terms", "model_id": "..."}}}
          ]
        }
      }
    }
  }
}
```

**Note:** In this configuration, the normalization processor in the search pipeline is bypassed. Scores from the hybrid sub-queries are combined as a raw sum (equivalent to a `bool` query with `should` clauses), and then the `function_score` functions are applied.

**Version:** 2.11+

kolchfa-aws:
Hi @martin-gaievski, this feature is for 3.6+, correct? Not 2.11+?

martin-gaievski:
> Hi [@martin-gaievski](https://github.com/martin-gaievski), this feature is for 3.6+, correct? Not 2.11+?

So I'm not sure how to put version correctly. Actual logic of hybrid not supporting nested HQ stays from 2.11, and in 3.6 I've added validation with user facing message.

kolchfa-aws:
[...]

PR #12088, opened 2026-03-11
Document hybrid query nesting limitations

martin-gaievski:
### Description
Updated hybrid query page with details of unsupported nesting patterns and available workaround.

### Issues Resolved
https://github.com/opensearch-project/documentation-website/issues/12079

### Version
3.6+

### Frontend features
N/A

### Checklist

### DIFF
--- a/_query-dsl/compound/hybrid.md
+++ b/_query-dsl/compound/hybrid.md
@@ -148,14 +148,13 @@ Starting with OpenSearch 3.5, you can use hybrid queries on indexes with more th
 
 ## Limitations
 
-The hybrid query is designed to be the top-level query in a search request. It cannot be nested inside other compound or wrapper queries such as `function_score`, `constant_score`, `script_score`, or `boosting`. This restriction also applies to multi-level nesting, for example, `bool` containing `function_score` containing `hybrid`. Nesting a hybrid query inside these wrapper queries may produce a runtime error or silently bypass the normalization pipeline.
-{: .important}
+Hybrid query is designed to be a top-level query in a search request. It cannot be nested inside other compound or wrapper queries such as `function_score`, `constant_score`, `script_score`, or `boosting`. This restriction also applies to multilevel nesting, for example, a `bool` query containing a `function_score` query that itself contains a `hybrid` query. Nesting a hybrid query inside these wrapper queries may produce a runtime error or silently bypass the normalization pipeline.
 
-The hybrid query uses a specialized scoring mechanism that is incompatible with wrapper queries. These wrapper queries use a different internal scorer that bypasses the hybrid query's per-subquery score collection, which is required for the normalization and combination pipeline to function correctly.
+Hybrid query uses a specialized scoring mechanism that is incompatible with wrapper queries. Wrapper queries use a different internal scorer that bypasses the hybrid query's per-subquery score collection, which is required for the normalization and combination pipeline to function correctly.
 
-If you need to apply score boosting functions to hybrid search results, replace the `hybrid` clause with a `bool` query using `should` clauses containing the same subqueries. This workaround works on all OpenSearch versions that support hybrid query:
+To apply score-boosting functions to hybrid search results, replace the `hybrid` query with a `bool` query and move your subqueries into `should` clauses. This alternative works in all OpenSearch versions that support hybrid queries.
 
-Instead of the following **unsupported** query:
+For example, the following query is unsupported:
 
 ```json
 GET /my-index/_search?search_pipeline=my-pipeline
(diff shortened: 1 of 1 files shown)


_query-dsl/compound/hybrid.md
```diff
@@ -146,6 +146,61 @@ Starting with version 3.5, the [`min_score`]({{site.url}}{{site.baseurl}}/api-re
 Starting with OpenSearch 3.5, you can use hybrid queries on indexes with more than 512 shards. OpenSearch automatically disables batched reduction to ensure proper score normalization across all shards. No configuration is required. Note that memory usage on the coordinating node may be higher for indexes with a large number of shards. The `_msearch` endpoint does not support automatic handling of batched reduction. For multi-search requests with hybrid queries across many shards, use the `_search` endpoint with index patterns or aliases instead.
 {: .note}
 
+## Limitations
+
+The hybrid query is designed to be the top-level query in a search request. It cannot be nested inside other compound or wrapper queries such as `function_score`, `constant_score`, `script_score`, or `boosting`. Attempting to do so returns an error. This restriction also applies to multi-level nesting, for example, `bool` containing `function_score` containing `hybrid`. Starting with version 3.6, validation coverage was extended to detect these unsupported nesting patterns and return a clear error message.
```
heemin32: I think we can omit the last sentence and make this document available for all version because this limitation is true for all version.

```suggestion
The hybrid query is designed to be the top-level query in a search request. It cannot be nested inside other compound or wrapper queries such as `function_score`, `constant_score`, `script_score`, or `boosting`. Attempting to do so returns an error. This restriction also applies to multi-level nesting, for example, `bool` containing `function_score` containing `hybrid`.
```

heemin32:
I think this update should be applied to all versions.

_query-dsl/compound/hybrid.md
martin-gaievski: In 3.5 and earlier versions we only validate for bool query, it's possible to pass hybrid inside function, script and constant, same for nested hybrid at level 2 and lower. It would be technically not correct to state validation is there, we cannot backport it, and user may read this and conclude: if I'm not getting validation error my query is correct.

heemin32: My concern is the other way around. By having this document only for 3.6+ and specifying `Starting with version 3.6,` customer might think it is not working starting 3.6 only.

martin-gaievski: Ok, let me put a more relaxed definition for "in earlier versions". The actual history is complex, and if we put all these details into the documentation it would make the docs harder to read for users who just want to know "don't nest hybrid, use bool+should instead":
- hybrid query initially released in 2.11, nested HQ not blocked, workaround is possible
- in 2.12 we blocked nested hybrid query for bool queries only
- in 2.14 we fixed incorrectly blocked queries for the scenario when a query is rewritten to bool by OpenSearch itself (e.g., index alias with filter), technically hybrid was the top query, but the blocker added in 2.12 stopped that query with the same validation error
- in 2.19 we added a block for disjunction query
- in 3.1 we added a blocker for hybrid query nested into another hybrid query

_query-dsl/compound/hybrid.md
```diff
@@ -146,6 +146,61 @@ Starting with version 3.5, the [`min_score`]({{site.url}}{{site.baseurl}}/api-re
...
 {: .note}
 
+## Limitations
+
+The hybrid query is designed to be the top-level query in a search request. It cannot be nested inside other compound or wrapper queries such as `function_score`, `constant_score`, `script_score`, or `boosting`. This restriction also applies to multi-level nesting, for example, `bool` containing `function_score` containing `hybrid`. Starting with version 3.6, validation coverage was extended to detect these unsupported nesting patterns and return a clear error message. In earlier versions, nesting a hybrid query inside these wrapper queries may produce a runtime error or silently bypass the normalization pipeline.
```
kolchfa-aws: @martin-gaievski Do you want to remove the "starting with version 3.6" so we can merge this PR and backport to the current version? Then we can include the 3.6 text in another PR.

martin-gaievski: I see, that's because 3.6 is not released yet. This makes sense, just pushed commit to address this

_query-dsl/compound/hybrid.md
```diff
@@ -146,6 +146,60 @@ Starting with version 3.5, the [`min_score`]({{site.url}}{{site.baseurl}}/api-re
...
+      "functions": [{"field_value_factor": {"field": "popularity"}}]
+    }
+  }
+}
+```
+
+Instead, use the following equivalent query:
+
+```json
+GET /my-index/_search
```
kolchfa-aws, line 181: @martin-gaievski Should this have `?search_pipeline=my-pipeline`?

martin-gaievski: search pipeline is optional, it doesn't have effect on results because we are changing hybrid query to bool query

PR #12155, opened 2026-03-26
Add include, exclude, and partition docs for terms aggregation

### LINKED ISSUE #12074: [DOC] Gaps in String Terms Aggregation Parameters
**What do you want to do?**
 
- [X] Add new documentation
- [ ] Other 

Figured out that for string terms aggregations, we don't have documentation for partitioning logic based on values, noticed few more params undocumented.

Ref: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/IncludeExclude.java

Gaps:
- include
- exclude
- partition 
- num_partitions

Not sure exactly where does the documentation needs to reside.

**Version:** all OpenSearch versions, this was inherited as it is from ES 7.10 to OpenSearch 1.0 during the fork.
 
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/IncludeExclude.java

kovan:
## Description

Document the `include`, `exclude`, `partition`, and `num_partitions` parameters for the terms aggregation, which were completely undocumented.

Three filtering modes are covered:
- **Regular expression filtering**: `include` and `exclude` accept regex strings (Lucene syntax, max 1000 chars by default)
- **Exact value filtering**: `include` and `exclude` accept arrays of exact values
- **Partition-based filtering**: `include` accepts an object with `partition` and `num_partitions` for iterating over high-cardinality fields

All information verified against the source code at `IncludeExclude.java`.

## Issues Resolved

Resolves #12074

## Check List

- [x] New functionality includes documentation
- [x] Commits are signed per the DCO using `--signoff`

### DIFF
--- a/_aggregations/bucket/terms.md
+++ b/_aggregations/bucket/terms.md
@@ -97,16 +97,16 @@ Separately, the `shard_min_doc_count` parameter is used to filter out the unique
 
 When using concurrent segment search, the `shard_min_doc_count` parameter is not applied to each segment slice. For more information, see the [related GitHub issue](https://github.com/opensearch-project/OpenSearch/issues/11847).
 
-## Filtering values with `include` and `exclude`
+## Filtering values to a subset
 
-You can use the `include` and `exclude` parameters to filter which term values appear in the aggregation buckets. When both are specified, `include` is evaluated first, and then `exclude` is applied to the result.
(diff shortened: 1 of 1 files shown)


kovan:
@kolchfa-aws This is ready for doc review. The content was verified against the source code in IncludeExclude.java. Happy to address any feedback.

kolchfa-aws (approved):
Thank you, @kovan!

PR #12321, opened 2026-04-23
Fix Java version mappings in installation guide

pawelw1:
Corrected Java version mappings for OpenSearch 3.2.0 and 3.5.0.

### Description
OpenSearch 3.2.0 to 3.4.0 is shipped with Java 24, and 3.5.0 to 3.6.0 with Java 25. The documentation states that OpenSearch is shipped only with compatible Java versions. However, current documentation still states that 3.2.0+ and 3.5.0+ are compatible only with Java 21. 

### Version
3.2.0, 3.3.x., 3.4.0, 3.5.0 and 3.6.0 

### Checklist


kolchfa-aws:
@peterzhuamazon Could you review this PR?

peterzhuamazon:
Hi @pawelw1 as you may notice those are minimum compat version which is always 21 for 2.19 and 3.x for now.
We are bundling 24 and 25 but min compat version is 21 until next time we change it. Therefore we can close this PR @kolchfa-aws .

Thanks.

bhupendra-mskhatri:
> Hi @pawelw1 as you may notice those are minimum compat version which is always 21 for 2.19 and 3.x for now. We are bundling 24 and 25 but min compat version is 21 until next time we change it. Therefore we can close this PR @kolchfa-aws .
> 
> Thanks.

Hi @peterzhuamazon 
https://docs.opensearch.org/latest/install-and-configure/install-opensearch/index/
The table in the above link lists compatible java versions for each opensearch version. 
For opensearch versions 1.x and 2.x the compatible java versions has multiple jdk versions. But 3.x lists just jdk 21. Should this not be updated to include 25 as well?

pawelw1:
> as you may notice those are minimum compat version

@peterzhuamazon  The documentation doesn't state that this is a minimal supported version.

AntonEliatra:
I would also agree that the documentation should be updated, feedback in the forum also confirms this. I would suggest to list the versions out same as it is done for `2.12.0+` for example.

peterzhuamazon:
adding @reta as he mainly update this section regarding Java version. Thanks.

_install-and-configure/install-opensearch/index.md
```diff
@@ -45,8 +45,8 @@ OpenSearch Version | Compatible Java Versions | Bundled Java Version
 1.3.x          | 8, 11, 14  | 11.0.25+9
 2.0.0--2.11.x    | 11, 17     | 17.0.2+8
 2.12.0+        | 11, 17, 21 | 21.0.5+11
-3.2.0+        | 21 | 24.0.2+12
-3.5.0+        | 21 | 25.0.2+10
+3.2.0+        | 24 | 24.0.2+12
```
reta, line 48:
```suggestion
3.2.0+        | 21, 24 | 24.0.2+12
```

_install-and-configure/install-opensearch/index.md
```diff
@@ -45,8 +45,8 @@ OpenSearch Version | Compatible Java Versions | Bundled Java Version
...
+3.5.0+        | 25 | 25.0.2+10
```
reta, line 49:
```suggestion
3.5.0+        | 21, 25 | 25.0.2+10
```

reta:
> adding @reta as he mainly update this section regarding Java version. Thanks.

Thanks @pawelw1 , there are two things to take into account:
 - 21 is the minimal baseline (OS 3.x could be run on)
 - although we test on every JDK, the compatible JDKs would be a combination of bundled one + all LTS versions >= baseline

Thank you

kolchfa-aws:
@pawelw1 Could you apply review comments to the PR?

kolchfa-aws:
Closing in favor of #12467

PR #12427, opened 2026-05-13
Update split index API documentation

kolchfa-aws:
Update split index API documentation with autogenerated endpoints, examples, explanations, and parameter tables.

### Checklist


_api-reference/index-apis/split.md
```diff
@@ -12,40 +12,146 @@ redirect_from:
...
+```
+{% include copy-curl.html %}
+
+Mappings cannot be specified in the split request. The target index inherits all mappings from the source index.
+{: .important}
+
+The split operation performs the following steps:
+
+1. Allocates a new target index with an identical mapping and configuration but a higher primary shard count.
+1. Establishes hard links from the source segments into the target index directory. If the file system lacks hard-link support, a full byte-level copy occurs instead, which takes significantly longer.
```
cwperks, line 59: nit: make sure the steps are numbered 1, 2, 3, 4

kolchfa-aws: Jekyll numbers lists automatically so numbering them 1., 1., 1. ensures that no number is missed in case steps change.

cwperks (approved):

PR #12544, opened 2026-06-19
Add hybrid search pre-filtering documentation

### LINKED ISSUE #1811: [FEATURE] Improve the validation error info for the hybrid query filter field
Yes. Hybrid query supports a common filter since 3.0 but only expect the filter to be a query object which means it cannot support the filter like 
```
"filter": [
        {
          "term": { "field1": "value1" }
        },
        {
          "term": { "field2": "value2" }
        }
]
```

We should provide better error info rather than simply saying “Field is not supported by [hybrid] query”.

Besides we should also improve our doc to include this info and guide users that they can use filter like
```
"filter": {
  "bool": {
    "must": [
      {"term": {"field1": "value1"}},
      {"term": {"field2": "value2"}}
    ]
  }
}
```

https://docs.opensearch.org/latest/query-dsl/compound/hybrid/
https://opensearch.org/blog/introducing-common-filter-support-for-hybrid-search-queries/

Or we can consider support the array format.

_Add any other context or screenshots about the feature request here._

RexyZhang2001:
### Description
Document the hybrid query common filter as pre-filtering: add a new pre-filtering page, clarify the filter parameter in the hybrid query reference, and cross-link the post-filtering page. The filter must be a single query object; multiple conditions are combined in a bool query.

### Issues Resolved
Closes https://github.com/opensearch-project/neural-search/issues/1811

### Version
3.0

### Frontend features

### Checklist

### DIFF
--- a/_query-dsl/compound/hybrid.md
+++ b/_query-dsl/compound/hybrid.md
@@ -22,7 +22,7 @@ The following table lists all top-level parameters supported by `hybrid` queries
 Parameter | Description
 :--- | :---
 `queries` | An array of one or more query clauses that are used to match documents. A document must match at least one query clause in order to be returned in the results. The documents' relevance scores from all query clauses are combined into one score by applying a [search pipeline]({{site.url}}{{site.baseurl}}/search-plugins/search-pipelines/index/). The maximum number of query clauses is 5. Required.
-`filter` | A filter to apply to all the subqueries of the hybrid query. 
+`filter` | A filter to apply to all the subqueries of the hybrid query. The filter must be a single query object. To apply multiple filter conditions, combine them in a [Boolean query]({{site.url}}{{site.baseurl}}/query-dsl/compound/bool/). For more information, see [Hybrid search with pre-filtering]({{site.url}}{{site.baseurl}}/vector-search/ai-search/hybrid-search/pre-filtering/).
 
 ### Rescoring hybrid queries
 Introduced 2.18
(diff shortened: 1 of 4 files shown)


heemin32 (approved):
Overall, LGTM. One suggestion: would it make more sense to combine `post-filtering.md` and `pre-filtering.md` into a single `filtering.md` document that covers both pre-filtering and post-filtering? I will leave it to @kolchfa-aws to decide.

heemin32:
@kolchfa-aws This is for 3.0 and above.

kolchfa-aws:
@RexyZhang2001 I think it makes sense to keep pre- and post-filtering pages separate. That said, could you add some information so the users can decide when to use pre-filtering and when to use post-filtering? It can go into a parent page (either create a new "Using filtering with hybrid search" parent page and make the pre- and post-filtering children of that page, or just add a section "Filtering data" in the current parent page). That way users can see the difference between the two options right away and we can help them decide which one to use. Thanks!

RexyZhang2001:
Hi @kolchfa-aws 
Thanks for the suggestion! I've added a new section to the current hybrid search parent page with a comparison table that summarizes the differences between pre-filtering and post-filtering. This should help users decide which approach fits their use case before reading both child pages first.

I would really appreciate it if you could take a look at this blog change as well when you get a chance, thanks: https://github.com/opensearch-project/project-website/pull/4169

heemin32 (approved):

kolchfa-aws (approved):
Thanks for the updates, @RexyZhang2001! Doc review complete and changes applied.

PR #12704, opened 2026-06-26
Remove /404.html redirect from index.md on 2.5

kolchfa-aws:
The /404.html redirect is now in _about/index.md. Having it in both index.md and _about/index.md causes a conflict where index.md wins, preventing the 404 page from redirecting to the about page.


peterzhuamazon (approved):
Thanks

PR #12720, opened 2026-06-30
Clarify function_score default match_all behavior

MdTanwer:
## Summary

- Documents that omitting the top-level `query` in `function_score` defaults to `match_all` with a base score of `1`
- Clarifies that function `filter`s affect scoring, not document inclusion
- Explains why documents matching no function filters still receive `_score: 1` with default settings
- Fixes misleading wording in the "Applying the scoring function to a subset of documents" section
- Adds a books example and `bool` query workaround for filtering by function filters

Resolves documentation gap identified in [opensearch-project/OpenSearch#21172](https://github.com/opensearch-project/OpenSearch/issues/21172).

### DIFF
--- a/_query-dsl/compound/function-score.md
+++ b/_query-dsl/compound/function-score.md
@@ -12,9 +12,15 @@ redirect_from:
 
 Use a `function_score` query if you need to alter the relevance scores of documents returned in the results. A `function_score` query defines a query and one or more functions that can be applied to all results or subsets of the results to recalculate their relevance scores.
 
+## Default query behavior
+
+If you omit the top-level `query` parameter, `function_score` defaults to `match_all`. Every document in the index is considered a match and receives a base query score of `1`.
+
+`function_score` changes how documents are **ranked**, not which documents are returned (unless you use `min_score`). To limit which documents are returned, provide an explicit top-level `query` or wrap `function_score` in a `bool` query.
+
 ## Using one scoring function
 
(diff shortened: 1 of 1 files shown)


kolchfa-aws:
@mgodwan Could you review this PR for technical accuracy?

_query-dsl/compound/function-score.md
```diff
@@ -686,6 +750,8 @@ The scores given by each function are combined using the `score_mode` parameter,
 - `max`: The maximum score is taken.
 - `min`: The minimum score is taken.
 
+If a document matches none of the function filters, the function score remains at the neutral value `1` (for `multiply`, `sum`, `avg`, `max`, and `min`). For `first`, no function score is applied and the factor also remains `1`.
```
kolchfa-aws: The second sentence reads as if `first` never applies a function. Looking at [the code](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/search/function/FunctionScoreQuery.java#L559), `factor` is initialized to `1d` which is a fallback for all score modes. The`first` [does apply the first matching function](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/search/function/FunctionScoreQuery.java#L561-L567) but leaves `factor` at `1 ` when nothing matches, exactly the same as every other mode. So `first` behaves in the same way and is not a special case.

kolchfa-aws (approved):
Thank you, @MdTanwer! While reviewing, I noticed some other missing information so filled it in as part of this PR. I also unified most of the examples on this page to use one index. All examples verified against an OpenSearch instance and all added claims verified against the source code.

PR #12723, opened 2026-06-30
Improve the workload group settings table and intro

dzane17:
### Description
Improves the **Supported settings** table and intro on the workload group settings page:
- Reduces the table to three columns (**Setting**, **Type**, **Description**) so the descriptions have room to breathe, folding the equivalent request parameter and cluster setting into the **Description** column as bold, `<br>`-separated lines (per maintainer feedback, keeping it a plain Markdown table).
- Deep-links the request parameters to the search API **Query parameters** section.
- Trims the intro and smooths the lead-in to the benefits list.

Before:
before

After:
after

### Issues Resolved
N/A

### Version
3.7+

### Checklist


kolchfa-aws:
@dzane17 Thank you for the PR. We usually avoid HTML tables unless absolutely necessary. Adding HTML tables with custom styles to documentation makes the docs harder to maintain and to contribute to. The theme handles responsive column sizing, which in this case is not great. In situations like these, we just combine the columns into one. How about making one column "Description", and putting Equivalent request parameter and Equivalent cluster setting in that same column after the description? I would make these bold and separate them into their own lines using `<br>`.

dzane17:
@kolchfa-aws I've reverted back to markdown table. Reducing to 4 columns still consumed a lot of horizontal space so I combined the last 2 columns into description. Let me know what you think. PR description contains the updated screenshot.

kolchfa-aws (approved):
Thank you, @dzane17! I added one more line to separate the equivalent parameters. LGTM

PR #12729, opened 2026-07-01
Add required permissions to the API pages

kolchfa-aws:
Add required permissions to the API documentation

### Checklist

### DIFF
--- a/_api-reference/alias/alias-exists.md
+++ b/_api-reference/alias/alias-exists.md
@@ -71,6 +71,10 @@ response = client.indices.exists_alias(
     python=step1_python %}
 <!-- spec_insert_end -->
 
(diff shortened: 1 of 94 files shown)


cwperks (approved):
@kolchfa-aws looks good and something that's long overdue :).

For cluster permissions (generally those that begin with cluster:, but with some notable exceptions) these are treated like all-or-nothing and are simple.

For indices: actions, these can sometimes be more complicated bc OpenSearch needs to resolve to concrete indices from the requests. Particularly with some actions around updating aliases that we may wish to elaborate on in the future.

Additionally some actions like `_cat/indices` ensure that the calling user has permission to get index settings across all non-hidden indices which can be subtle.

PR #12795, opened 2026-07-13
Document workload group rule matching and precedence

dzane17:
### Description
Documents how WLM auto-tagging rules match a request and how OpenSearch resolves precedence when multiple rules match. Adds a "Rule matching and precedence" section to the workload group rules page covering:

- Matching within a single rule: multiple attributes are AND'd; multiple values within one attribute (including the `principal` username/role subfields) are OR'd.
- Choosing between multiple matching rules: attribute priority, then match score, then no assignment on an unresolved tie.

Also clarifies on the rule-based auto-tagging page that when rules remain tied after tie-breaking, no value is assigned (OpenSearch does not choose arbitrarily).

### Issues Resolved
N/A

### Version
2.19 and later

### Checklist


kaushalmahi12 (approved):

kolchfa-aws (approved):
Thank you, @dzane17! Doc review complete and changes applied.

PR #12855, opened 2026-07-22
Update CSP configuration docs to reflect yml-based config

sumukhswamy:
## Summary

- Replaces the deprecated `applicationConfig` + `cspHandler` plugin approach with the current `csp.*` and `csp-report-only.*` yml settings
- Adds a settings reference table for all new CSP keys
- Adds a report-only mode section
- Moves the old plugin-based approach to a deprecated section for reference

## Context

The `cspHandler` pre-response handler was removed in 2.17 as part of the Dynamic Config Service migration. The old API still accepts values but no longer applies them to the CSP header. The current way to configure CSP is via `opensearch_dashboards.yml`.

## Test plan

- [ ] Review rendered markdown on GitHub
- [ ] Verify yml examples match `config/opensearch_dashboards.yml` in the OSD repo

### DIFF
--- a/_dashboards/csp/csp-dynamic-configuration.md
+++ b/_dashboards/csp/csp-dynamic-configuration.md
@@ -1,58 +1,108 @@
 ---
 layout: default
-title: Configuring CSP rules for frame ancestors
+title: Configuring CSP rules
 parent: Settings and administration
 nav_order: 60
 has_children: false
 ---
 
-# Configuring CSP rules for frame ancestors
+# Configuring CSP rules
 Introduced 2.13
 {: .label .label-purple }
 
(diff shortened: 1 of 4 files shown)


angle943 (approved):

kolchfa-aws (approved):
Thank you, @sumukhswamy! Doc review complete and changes applied.

_dashboards/csp/csp-dynamic-configuration.md
```diff
@@ -1,58 +1,82 @@
...
+csp-report-only.allowedConnectSources: ["https://api.example.com"]
+csp-report-only.allowedImgSources: ["https://cdn.example.com"]
+```
+{% include copy.html %}
 
-An example of the `Copy as cURL` option in Firefox is shown in the following image.
+## Configuring CSP rules using applicationConfig (deprecated)
 
-![Copying as cURL in Firefox]({{site.url}}{{site.baseurl}}/images/dashboards/copy-as-curl.png)
+**Deprecated.** In OpenSearch Dashboards 2.13--2.16, you could set the `frame-ancestors` directive dynamically using a REST API using the `applicationConfig` and `cspHandler` plugins. This approach is no longer functional. Use the `csp.*` settings instead.
```
kolchfa-aws:
```suggestion
**Deprecated.** In OpenSearch Dashboards 2.13--2.16, you could set the `frame-ancestors` directive dynamically through a REST API by using the `applicationConfig` and `cspHandler` plugins. This approach is no longer functional. Use the `csp.*` settings instead.
```

PR #12922, opened 2026-08-12
Docs/Add documentation for standalone audit logging settings

### LINKED ISSUE #6303: [FEATURE] Standalone audit logging for SSL-only mode
### Description
  
Enable audit logging as an independent capability in SSL-only mode (`plugins.security.ssl_only: true`) without requiring full security (FGAC/authentication/authorization).
  
### Background
  
Currently, audit logging only produces events when the full security stack is active. Users running OpenSearch with TLS-only (for log analytics, observability, internal clusters) have no audit trail. Compliance frameworks (SOC2, HIPAA, PCI-DSS) require audit logs regardless of access control configuration.
  
This is the first phase of a multi-phase effort toward a fully standalone audit plugin:
  
  1. **SSL-only mode** Only TLS encryption, no authentication or authorization 
  2. **Disabled mode** No security at all
  3. **Standalone audit plugin** (extracted from security plugin) — future
  
### Scope

- New `REQUEST_AUDIT` category for all REST/transport actions
- Client certificate identity enrichment (mTLS)
- Request body logging with sensitive header exclusion
- Index resolution and bulk per-item event tracking
- Ignore-users and ignore-requests filtering
- Transport-layer interception
- Document-level compliance tracking (COMPLIANCE_DOC_WRITE, COMPLIANCE_DOC_READ)
- All existing sinks supported (internal_opensearch, log4j, webhook, Kafka)
- 17 settings dynamically configurable via `PUT _cluster/settings`
  
### Configuration
  
  ```yaml
  plugins.security.ssl_only: true
  plugins.security.audit.type: log4j  # or internal_opensearch, webhook
  ```
  
Related
  
- Broader vision: https://github.com/opensearch-project/.github/issues/501 
- Upstream PR (unified disabled_categories): https://github.com/opensearch-project/security/pull/6271

### LINKED ISSUE #6340: [FEATURE] Standalone audit logging for disabled security mode
### Description

Enable audit logging as an independent capability in disabled mode (`plugins.security.disabled: true`) without requiring any security features — no TLS, no authentication, no authorization.

### Background

With SSL-only mode audit logging merged (#6304, #6321), users running OpenSearch with TLS-only have full audit visibility. However, users running with security completely disabled (common in development, log analytics, and internal observability clusters) still have no audit trail.

Compliance frameworks (SOC2, HIPAA, PCI-DSS, GDPR) require audit logs even when access control is not configured.

This is the second phase of the multi-phase standalone audit effort:
1. SSL-only mode (#6304)
2. **Disabled mode** ← This issue
3. Standalone audit plugin (extracted from security plugin) — future

### Scope

All features from SSL-only mode now work in disabled mode:
- `REQUEST_AUDIT` category for all REST/transport actions
- Request body logging with sensitive header exclusion
- Index resolution and bulk per-item event tracking
- Ignore-users and ignore-requests filtering
- Transport-layer interception (`TRANSPORT_AUDIT`)
- Document-level compliance tracking (`COMPLIANCE_DOC_WRITE`, `COMPLIANCE_DOC_READ`)
- All existing sinks supported (internal_opensearch, log4j, webhook, Kafka)
- 17 settings dynamically configurable via `PUT _cluster/settings`

### What's different from SSL-only mode

| | SSL-only | Disabled |
|---|---|---|
| TLS on REST | Required | Not required (plain HTTP) |
| TLS on Transport | Required | Not required |
| Client cert identity (mTLS) | Available | Not available (no TLS) |
| User identity in events | Source IP + cert CN | Source IP only |

### Configuration

```yaml
plugins.security.disabled: true
plugins.security.audit.type: log4j  # or internal_opensearch, webhook
```

### Related

- SSL-only audit logging: #6304 
- PR review follow-up: #6321 
- Broader vision: [PROPOSAL] Standalone Audit Logging Plugin for OpenSearch: [#501](https://github.com/opensearch-project/.github/issues/501)
- Unified disabled_categories: #6271

Taiwo435:
### Description

Add comprehensive documentation for the standalone audit logging feature and all associated cluster settings changes introduced across multiple Security plugin PRs.

**Changes:**

1. **`_security/audit-logs/index.md`** (updated):
   - Add intro explaining two audit modes (standard vs standalone)
   - Add new tracked events: `REQUEST_AUDIT`, `TRANSPORT_AUDIT`, `RESOURCE_ACCESS_GRANTED`, `RESOURCE_ACCESS_DENIED`, `RESOURCE_SHARING_CHANGED`
   - Document unified `disabled_categories` setting with deprecation warning for split settings
   - Document `body_logging_exclusions` with action groups (full section: pattern types, group naming, bulk behavior, interaction with `log_request_body`, examples)
   - Document `log4j.enable_mdc_routing` setting
   - Document dynamic `audit.enabled` toggle via cluster settings API

2. **`_security/audit-logs/standalone.md`** (new page):
   - Explains standalone audit mode (works in SSL-only and security-disabled modes without FGAC)
   - Requirements (`enable_standalone: true` + `audit.type`)
   - Supported sinks, tracked events, dynamic configuration
   - Compliance tracking (document write/read)
   - Example configurations for both SSL-only and disabled-security modes

3. **`_install-and-configure/configuring-opensearch/security-settings.md`** (updated):
   - Add `body_logging_exclusions`, `action_groups.<NAME>`, `log4j.enable_mdc_routing`, `enable_standalone` settings
   - Add note about `Setting.Property.Sensitive` on all audit/compliance settings

### Issues Resolved

Closes opensearch-project/security#6303
Closes opensearch-project/security#6340

Related Security plugin PRs:
- opensearch-project/security#6304 — Standalone audit logging for SSL-only mode
- opensearch-project/security#6321 — PR review feedback (Sensitive property)
- opensearch-project/security#6341 — Standalone audit in disabled security mode
- opensearch-project/security#6347 — Body logging exclusions with action groups
- opensearch-project/security#6349 — Log4j MDC routing
- opensearch-project/security#6368 — Explicit opt-in with `enable_standalone`
- opensearch-project/security#6271 — Unified `disabled_categories` setting
- opensearch-project/security#6352 — Resource sharing audit categories

### Version

3.8

### Frontend features

N/A - these are backend cluster settings, not Dashboards UI features.

### Checklist

### DIFF
--- a/_install-and-configure/configuring-opensearch/security-settings.md
+++ b/_install-and-configure/configuring-opensearch/security-settings.md
@@ -245,7 +245,7 @@ The Security plugin supports the following audit log settings:
 
 - `opendistro_security.audit.config.disabled_transport_categories` (Dynamic): A list of transport layer categories to be ignored by the logger. Valid values are `AUTHENTICATED` and `GRANTED_PRIVILEGES`.
 
-All `plugins.security.audit.config.*` and `plugins.security.audit.compliance.*` settings are marked with `Setting.Property.Sensitive`. This means only security admin users can view or modify their values via the cluster settings API; values are masked for non-admin users.
+The dynamic audit filter settings and all `plugins.security.audit.compliance.*` settings are marked with `Setting.Property.Sensitive`. This means only security admin users can view or modify their values via the cluster settings API; values are masked for non-admin users. The `plugins.security.audit.config.body_logging_exclusions` and `plugins.security.audit.config.action_groups.<NAME>` settings are exceptions and are not marked sensitive.
 
 ## Hostname verification and DNS lookup settings
 
--- a/_install-and-configure/configuring-opensearch/security-settings.md
+++ b/_install-and-configure/configuring-opensearch/security-settings.md
@@ -205,6 +205,14 @@ The Security plugin supports the following audit log settings:
 
 - `plugins.security.audit.config.cert_alias` (Static): An alias to the certificate used for audit log access.
 
+- `plugins.security.audit.config.body_logging_exclusions` (Dynamic): A list of action group names or raw action/path patterns for which request body logging is suppressed. Default is `[]` (empty, log all bodies). See [Body logging exclusions]({{site.url}}{{site.baseurl}}/security/audit-logs/index/#body-logging-exclusions) for details.
+
+- `plugins.security.audit.config.action_groups.<NAME>` (Static): Defines named groups of action/path patterns for use with `body_logging_exclusions`. Each group is a comma-separated string of transport action patterns and/or REST paths. Supports wildcards.
+
+- `plugins.security.audit.config.log4j.enable_mdc_routing` (Static): Enables MDC (Mapped Diagnostic Context) routing for the Log4j audit sink. When enabled, audit events set MDC keys (`audit_category`, `audit_action`, `audit_user`, `audit_request_type`) that can be used by Log4j routing appenders. Default is `false`.
+
+- `plugins.security.audit.enable_standalone` (Static): Enables standalone audit logging mode for clusters running without fine-grained access control (SSL-only or security-disabled modes). Must be set to `true` alongside `plugins.security.audit.type` to activate standalone audit. Default is `false`.
+
 - `plugins.security.audit.config.pemkey_filepath` (Static): The `/config` relative file path to the Privacy Enhanced Mail (PEM) key used for audit logging.
 
 - `plugins.security.audit.config.pemkey_content` (Static): The Base64-encoded content of the PEM key used for audit logging. This is an alternative to `...config.pemkey_filepath`.
@@ -236,6 +244,8 @@ The Security plugin supports the following audit log settings:
 - `opendistro_security.audit.config.disabled_rest_categories` (Dynamic): A list of REST categories to be ignored by the logger. Valid values are `AUTHENTICATED` and `GRANTED_PRIVILEGES`.
 
 - `opendistro_security.audit.config.disabled_transport_categories` (Dynamic): A list of transport layer categories to be ignored by the logger. Valid values are `AUTHENTICATED` and `GRANTED_PRIVILEGES`.
+
+All `plugins.security.audit.config.*` and `plugins.security.audit.compliance.*` settings are marked with `Setting.Property.Sensitive`. This means only security admin users can view or modify their values via the cluster settings API; values are masked for non-admin users.
 
 ## Hostname verification and DNS lookup settings
 
--- a/_security/audit-logs/index.md
+++ b/_security/audit-logs/index.md
@@ -26,6 +26,13 @@ redirect_from:
 
 Audit logs let you track access to your OpenSearch cluster and are useful for compliance purposes or in the aftermath of a security breach. You can configure the categories to be logged, the detail level of the logged messages, and where to store the logs.
 
+OpenSearch supports two audit logging modes:
+
+- **Standard mode** (default): Requires the Security plugin with fine-grained access control (FGAC) enabled. Configuration is managed through the `audit.yml` file in the security index and through OpenSearch Dashboards.
+- **Standalone mode**: Works without FGAC---in SSL-only mode (`plugins.security.ssl_only: true`) or with security disabled (`plugins.security.disabled: true`). Configuration is managed entirely through `opensearch.yml` and dynamic cluster settings. See [Standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) for details.
+
+### Enabling audit logging (standard mode)
+
 Audit logging is disabled by default. To enable audit logging:
 
 1. Add the following line to `opensearch.yml` on each node:
@@ -59,6 +66,11 @@ Event | Logged on REST | Logged on transport | Description
 `BAD_HEADERS` | Yes | Yes | An attempt was made to spoof a request to OpenSearch with the Security plugin internal headers.
 `CLUSTER_SETTINGS_CHANGED` | No | Yes | A persistent or transient cluster setting was changed. Disabled by default.
 `INDEX_SETTINGS_CHANGED` | No | Yes | An index setting was changed. Disabled by default.
+`REQUEST_AUDIT` | Yes | No | A REST-layer request was received and processed. Generated in [standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) mode only.
+`TRANSPORT_AUDIT` | No | Yes | A transport-layer request was received on a node. Generated in [standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) mode only.
+`RESOURCE_ACCESS_GRANTED` | No | Yes | Access to a shared resource was granted. Disabled by default.
+`RESOURCE_ACCESS_DENIED` | No | Yes | Access to a shared resource was denied. Disabled by default.
+`RESOURCE_SHARING_CHANGED` | No | Yes | A resource sharing configuration was changed. Disabled by default.
 
 
 ## Audit log settings
@@ -260,11 +272,30 @@ config:
 
 ### Settings in opensearch.yml
 
-The following settings are stored in the `opensearch.yml` file.
+The following settings are stored in the `opensearch.yml` file. All audit settings with the `plugins.security.audit.config` prefix are dynamic---they can be changed at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/) without a node restart. Audit settings are marked as sensitive, meaning only security admin users can view or modify them via the cluster settings API.
+
+#### Enable or disable audit logging
+
+```yml
+plugins.security.audit.enabled: true
+```
+{% include copy.html %}
+
+Enables or disables audit logging globally. Default is `true`. This setting is dynamic and can be toggled at runtime:
+
+```json
+PUT _cluster/settings
+{
+  "persistent": {
+    "plugins.security.audit.enabled": false
+  }
+}
+```
+{% include copy.html %}
 
 #### Exclude categories
 
-You can configure disabled categories in `opensearch.yml` using the `plugins.security.audit.config` prefix. This is useful for non-fine-grained access control (FGAC) modes (SSL-only or security-disabled) for which the `audit.yml` security index is not available:
+You can configure disabled categories in `opensearch.yml` using the `plugins.security.audit.config` prefix. This is useful for non-FGAC modes (SSL-only or security-disabled) where the `audit.yml` security index is not available:
 
 ```yml
 plugins.security.audit.config.disabled_categories:
@@ -273,8 +304,8 @@ plugins.security.audit.config.disabled_categories:
 ```
 {% include copy.html %}
 
-The layer-specific settings (`disabled_rest_categories` and `disabled_transport_categories`) may be deprecated in a future version. Use the unified `disabled_categories` setting instead.
 {: .warning}
+The following layer-specific settings (`disabled_rest_categories` and `disabled_transport_categories`) are on a deprecation path and could be removed in a future release. Use the unified `disabled_categories` setting instead.
 
 The layer-specific settings are also available:
 
@@ -288,7 +319,147 @@ plugins.security.audit.config.disabled_transport_categories:
 ```
 {% include copy.html %}
 
-When both `disabled_categories` and the layer-specific settings are configured, a category is disabled on a given layer if it appears in either setting.
+When both `disabled_categories` and the layer-specific settings are configured, they work in tandem---a category is disabled on a given layer if it appears in either setting. A deprecation warning is logged when both are configured, encouraging migration to `disabled_categories` only.
+
+#### Body logging exclusions
+
+Request body logging is valuable for compliance but can be expensive at scale. For example, a cluster performing 100,000+ bulk writes per second generates enormous audit volume---most of it redundant data payloads. The `log_request_body: false` toggle is all-or-nothing: it either logs all bodies or none.
+
+Body logging exclusions give operators granular control: suppress request bodies for high-volume operations (like bulk ingestion) while still capturing bodies for searches, index creation, and admin operations that matter for investigation.
+
+##### How it works
+
+Each audit event's action name and REST path are matched against a set of exclusion patterns. When a match is found, the request body field is omitted from the audit event---all other fields (user, IP address, indices, timestamp, etc.) are preserved.
+
+Matching uses two identifiers per request:
+
+- **Transport action** --- The internal action name (for example, `indices:data/write/bulk[s][p]`). This is matched for transport-layer audit events.
+- **REST path** --- The HTTP request path (for example, `/_bulk`). This is matched for REST-layer audit events. REST paths always start with `/`.
+
+Both identifiers support wildcard patterns using `*` (for example, `indices:data/write/bulk*` matches `indices:data/write/bulk[s][p]`).
+
+##### Configuring action groups
+
+Action groups are named collections of action patterns and/or REST paths, defined statically in `opensearch.yml`. The group name is user-chosen---pick any name that is meaningful for your operations:
+
+```yml
+plugins.security.audit.config.action_groups.BULK: "indices:data/write/bulk*,/_bulk"
+plugins.security.audit.config.action_groups.SEARCH: "indices:data/read/search*,/_search"
+plugins.security.audit.config.action_groups.INDEX_ADMIN: "indices:admin/*"
+```
+{% include copy.html %}
+
+Each action group maps a name to a comma-separated list of patterns. Patterns can be:
+
+- Transport action patterns (contain `:`) --- for example, `indices:data/write/bulk*`
+- REST path patterns (start with `/`) --- for example, `/_bulk`
+- Wildcard patterns (contain `*`) --- for example, `indices:data/write/*`
+
+Action groups are static settings and require a node restart to change. The group names themselves are case-sensitive.
+
+##### Configuring body logging exclusions
+
+The `body_logging_exclusions` setting references action group names or raw patterns. It is dynamic and can be changed at runtime:
+
+```yml
+plugins.security.audit.config.body_logging_exclusions:
+  - BULK
+```
+{% include copy.html %}
+
+Or update at runtime without a restart:
+
+```json
+PUT _cluster/settings
+{
+  "persistent": {
+    "plugins.security.audit.config.body_logging_exclusions": ["BULK", "SEARCH"]
+  }
+}
+```
+{% include copy.html %}
+
+Each entry in the list is processed as follows:
+
+1. If the entry matches a defined action group name, that group's patterns are expanded.
+2. If it does not match any group name, the entry is treated as a raw pattern (literal or wildcard).
+
+A warning is logged for entries that don't look like an action pattern (no `:`), REST path (no `/`), or wildcard (no `*`), since they are unlikely to match any action.
+
+##### Bulk request behavior
+
+When `resolve_bulk_requests: true` is configured (logging individual bulk sub-items), the exclusion check uses the parent bulk action string. This means that excluding the `BULK` group suppresses bodies for **all** sub-items (index, update, delete) within that bulk request. You cannot selectively keep index-item bodies while dropping delete-item bodies within the same bulk request.
+
+##### Interaction with `log_request_body`
+
+Body logging exclusions only apply when `log_request_body` is `true`. If `log_request_body` is `false`, no bodies are logged regardless of the exclusion configuration.
+
+##### Example: Complete configuration
+
+```yml
+# opensearch.yml
+
+# Define action groups (static, requires restart)
+plugins.security.audit.config.action_groups.BULK: "indices:data/write/bulk*,/_bulk"
+plugins.security.audit.config.action_groups.SEARCH: "indices:data/read/search*,/_search"
+plugins.security.audit.config.action_groups.MONITORING: "cluster:monitor/*,indices:monitor/*"
+
+# Initial exclusions (can be updated at runtime via _cluster/settings)
+plugins.security.audit.config.body_logging_exclusions:
+  - BULK
+```
+{% include copy.html %}
+
+With this configuration:
+- Bulk write requests: body is **not** logged (excluded)
+- Search requests: body **is** logged (not excluded)
+- Index creation: body **is** logged (not excluded)
+- Monitoring requests: body **is** logged (not excluded, unless you add `MONITORING` to exclusions)
+
+To add search exclusions at runtime:
+
+```json
+PUT _cluster/settings
+{
+  "persistent": {
+    "plugins.security.audit.config.body_logging_exclusions": ["BULK", "SEARCH"]
+  }
+}
+```
+{% include copy.html %}
+
+To clear all exclusions (resume logging all bodies):
+
+```json
+PUT _cluster/settings
+{
+  "persistent": {
+    "plugins.security.audit.config.body_logging_exclusions": []
+  }
+}
+```
+{% include copy.html %}
+
+#### Log4j MDC routing
+
+When using the `log4j` audit sink, you can enable MDC (Mapped Diagnostic Context) routing to allow Log4j to route audit events to different appenders based on event properties:
+
+```yml
+plugins.security.audit.config.log4j.enable_mdc_routing: true
+```
+{% include copy.html %}
+
+When enabled, the following MDC keys are set on each audit event:
+
+- `audit_category` --- The audit event category (for example, `REQUEST_AUDIT`, `GRANTED_PRIVILEGES`)
+- `audit_action` --- The action name
+- `audit_user` --- The effective user
+- `audit_request_type` --- The request type
+
+This allows you to configure Log4j routing appenders in your `log4j2.properties` to split audit logs by category, user, or any other MDC key.
+
+This is a static setting and requires a node restart.
+{: .note}
 
 
 #### Configure the audit log index name
--- a/_security/audit-logs/index.md
+++ b/_security/audit-logs/index.md
@@ -28,7 +28,7 @@ Audit logs let you track access to your OpenSearch cluster and are useful for co
 
 OpenSearch supports two audit logging modes:
 
-- **Standard mode** (default): Requires the Security plugin with fine-grained access control (FGAC) enabled. Configuration is managed through the `audit.yml` file in the security index and through OpenSearch Dashboards.
+- **Standard mode** (default): Requires the Security plugin with fine-grained access control (FGAC) enabled. Configuration is managed through the `audit.yml` file in the security index and the REST API.
 - **Standalone mode**: Works without FGAC---in SSL-only mode (`plugins.security.ssl_only: true`) or with security disabled (`plugins.security.disabled: true`). Configuration is managed entirely through `opensearch.yml` and dynamic cluster settings. See [Standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) for details.
 
 ### Enabling audit logging (standard mode)
@@ -46,9 +46,7 @@ Audit logging is disabled by default. To enable audit logging:
 
 2. Restart each node.
 
-After this initial setup, you can use OpenSearch Dashboards to manage your audit log categories and other settings. In OpenSearch Dashboards, select **Security** and then **Audit logs**. 
-
-An alternative is to specify initial settings for audit logging in the `audit.yml` and `opensearch.yml` files (which file depends on the setting---see [Audit log settings](#audit-log-settings)). Thereafter, you can use Dashboards or the [Audit logs]({{site.url}}{{site.baseurl}}/security/access-control/api/#audit-logs) API to manage and update settings.
+After this initial setup, you can specify settings for audit logging in the `audit.yml` and `opensearch.yml` files (which file depends on the setting---see [Audit log settings](#audit-log-settings)). You can also use the [Audit logs]({{site.url}}{{site.baseurl}}/security/access-control/api/#audit-logs) API to manage and update settings.
 
 
 ## Tracked events
@@ -272,7 +270,7 @@ config:
 
 ### Settings in opensearch.yml
 
-The following settings are stored in the `opensearch.yml` file. All audit settings with the `plugins.security.audit.config` prefix are dynamic---they can be changed at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/) without a node restart. Audit settings are marked as sensitive, meaning only security admin users can view or modify them via the cluster settings API.
+The following settings are stored in the `opensearch.yml` file. The audit filter and compliance settings---for example, `log_request_body`, `resolve_indices`, `disabled_categories`, and the `plugins.security.audit.compliance.*` settings---are dynamic: they can be changed at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/) without a node restart. Most of these dynamic settings are also marked as sensitive, meaning only security admin users can view or modify them via the cluster settings API (`body_logging_exclusions` and `action_groups.<NAME>` are exceptions and are not sensitive). Other audit settings---such as `action_groups.<NAME>`, `log4j.enable_mdc_routing`, `config.index`, the thread pool settings, and the sink connection settings---are static and require a node restart.
 
 #### Enable or disable audit logging
 
@@ -304,8 +302,8 @@ plugins.security.audit.config.disabled_categories:
 ```
 {% include copy.html %}
 
-{: .warning}
 The following layer-specific settings (`disabled_rest_categories` and `disabled_transport_categories`) are on a deprecation path and could be removed in a future release. Use the unified `disabled_categories` setting instead.
+{: .warning}
 
 The layer-specific settings are also available:
 
@@ -320,6 +318,9 @@ plugins.security.audit.config.disabled_transport_categories:
 {% include copy.html %}
 
 When both `disabled_categories` and the layer-specific settings are configured, they work in tandem---a category is disabled on a given layer if it appears in either setting. A deprecation warning is logged when both are configured, encouraging migration to `disabled_categories` only.
+
+The `disabled_categories` settings suppress only REST and transport categories. They do not affect `COMPLIANCE_*` categories, which are governed solely by the compliance settings (`compliance.enabled`, the watched indices/fields, and the compliance ignore-users settings).
+{: .note}
 
 #### Body logging exclusions
 
@@ -464,7 +465,7 @@ This is a static setting and requires a node restart.
 
 #### Configure the audit log index name
 
-By default, the Security plugin stores audit events in a daily rolling index named `auditlog-YYYY.MM.dd`:
+By default, the Security plugin stores audit events in a daily rolling index named `security-auditlog-YYYY.MM.dd`:
 
 ```yml
 plugins.security.audit.config.index: myauditlogindex
@@ -499,7 +500,7 @@ plugins.security.audit.config.threadpool.max_queue_len: 100000
 
 ## Disabling audit logs
 
-To disable audit logs after they've been enabled, remove the `plugins.security.audit.type: internal_opensearch` setting from `opensearch.yml`, or switch off the **Enable audit logging** check box in OpenSearch Dashboards.
+To disable audit logs after they've been enabled, remove the `plugins.security.audit.type: internal_opensearch` setting from `opensearch.yml`, or set `plugins.security.audit.enabled` to `false` via the cluster settings API.
 
 ## Audit user account manipulation
 
--- a/_security/audit-logs/standalone.md
+++ b/_security/audit-logs/standalone.md
@@ -22,6 +22,8 @@ nav_order: 130
 
 Standalone audit logging enables audit logging for OpenSearch clusters that do not use fine-grained access control (FGAC). This includes clusters running in SSL-only mode (`plugins.security.ssl_only: true`) or with security disabled (`plugins.security.disabled: true`).
 
+In this context, *standalone* means audit logging is independent of fine-grained access control---not a separate plugin or process. It runs inside the Security plugin and reuses the same audit pipeline as standard mode; it just does not depend on authentication or authorization to produce events.
+
 Many organizations need audit trails for compliance (SOC2, HIPAA, PCI-DSS, GDPR) even when they don't require authentication or authorization. Standalone audit logging answers the question "who did what, when?" without requiring the full security infrastructure.
 
 {: .note }
--- a/_security/audit-logs/standalone.md
+++ b/_security/audit-logs/standalone.md
@@ -104,7 +104,9 @@ Identity information varies by security mode:
 
 ## Configuration
 
-All standalone audit settings are configured in `opensearch.yml` for initial values and can be dynamically updated at runtime using the cluster settings API. No security index is required.
+All standalone audit settings are configured in `opensearch.yml` for initial values and can be dynamically updated at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/). No security index is required.
+
+Standalone mode does not use the [Audit logs]({{site.url}}{{site.baseurl}}/security/access-control/api/#audit-logs) REST API or `audit.yml`---those manage the security index and require fine-grained access control. In standalone mode, use the Cluster settings API instead.
 
 ### Dynamic configuration
 
--- a/_security/audit-logs/standalone.md
+++ b/_security/audit-logs/standalone.md
@@ -61,7 +61,7 @@ For sink-specific configuration options, see [Audit log storage types]({{site.ur
 
 ## Tracked events
 
-Standalone audit logging introduces two event categories designed for environments without authentication:
+Standalone audit logging introduces two request-tracking categories designed for environments without authentication:
 
 Category | Layer | Description
 :--- | :--- | :---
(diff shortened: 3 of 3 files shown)


kolchfa-aws:
@cwperks Could you please review this PR?

_install-and-configure/configuring-opensearch/security-settings.md
```diff
@@ -237,6 +245,8 @@ The Security plugin supports the following audit log settings:
 
 - `opendistro_security.audit.config.disabled_transport_categories` (Dynamic): A list of transport layer categories to be ignored by the logger. Valid values are `AUTHENTICATED` and `GRANTED_PRIVILEGES`.
 
+The dynamic audit filter settings and all `plugins.security.audit.compliance.*` settings are marked with `Setting.Property.Sensitive`. This means only security admin users can view or modify their values via the cluster settings API; values are masked for non-admin users. The `plugins.security.audit.config.body_logging_exclusions` and `plugins.security.audit.config.action_groups.<NAME>` settings are exceptions and are not marked sensitive.
```
DarshitChanpura, line 248: This needs #6392 for the SSL-only read path, so it can't merge before that.

The role wording is incorrect regardless of #6392. The settings filter isn't role-aware. After #6392, SSL-only returns the non-secret config to any caller (no roles exist there), and FGAC keeps the whole `audit.*` subtree filtered for everyone. There's no "admins can view, non-admins masked" behavior.

Suggest something like: "These settings are registered `Setting.Property.Sensitive`, which keeps their values out of diagnostics and logs. In standalone SSL-only mode the non-secret dynamic config (`…config.*` and `…compliance.*`) is readable via `GET _cluster/settings`; credential-bearing sink settings (endpoint/route credentials, PEM keys, salt) stay hidden. `body_logging_exclusions` and `action_groups.<NAME>` aren't marked sensitive."

_security/audit-logs/index.md
```diff
@@ -260,11 +270,30 @@ config:
 
 ### Settings in opensearch.yml
 
-The following settings are stored in the `opensearch.yml` file.
+The following settings are stored in the `opensearch.yml` file. The audit filter and compliance settings---for example, `log_request_body`, `resolve_indices`, `disabled_categories`, and the `plugins.security.audit.compliance.*` settings---are dynamic: they can be changed at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/) without a node restart. Most of these dynamic settings are also marked as sensitive, meaning only security admin users can view or modify them via the cluster settings API (`body_logging_exclusions` and `action_groups.<NAME>` are exceptions and are not sensitive). Other audit settings---such as `action_groups.<NAME>`, `log4j.enable_mdc_routing`, `config.index`, the thread pool settings, and the sink connection settings---are static and require a node restart.
```
DarshitChanpura, line 273: Same as the security-settings.md note: needs #6392, and "only security admin users can view or modify them via the cluster settings API" is incorrect in either mode (SSL-only is readable by any caller; FGAC is filtered for everyone). Reword to describe `Property.Sensitive` instead of role-based masking.

Also, `action_groups.<NAME>` is listed here as a dynamic setting but it's static (`groupSetting(..., NodeScope)`) — the next sentence already says so. Drop it from this parenthetical; only `body_logging_exclusions` is a dynamic non-sensitive exception.

_security/audit-logs/index.md
```diff
@@ -59,6 +64,11 @@ Event | Logged on REST | Logged on transport | Description
 `BAD_HEADERS` | Yes | Yes | An attempt was made to spoof a request to OpenSearch with the Security plugin internal headers.
 `CLUSTER_SETTINGS_CHANGED` | No | Yes | A persistent or transient cluster setting was changed. Disabled by default.
 `INDEX_SETTINGS_CHANGED` | No | Yes | An index setting was changed. Disabled by default.
+`REQUEST_AUDIT` | Yes | No | A REST-layer request was received and processed. Generated in [standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) mode only.
```
DarshitChanpura, line 67: `REQUEST_AUDIT` has `origin=REST` but `request_layer=TRANSPORT` internally and can be suppressed via `disabled_transport_categories`. "No" is fine for a REST-origin event, but a footnote would be more accurate. Non-blocking.

_security/audit-logs/index.md
```diff
@@ -288,12 +317,155 @@ plugins.security.audit.config.disabled_transport_categories:
...
+
+#### Body logging exclusions
+
+Request body logging is valuable for compliance but can be expensive at scale. For example, a cluster performing 100,000+ bulk writes per second generates enormous audit volume---most of it redundant data payloads. The `log_request_body: false` toggle is all-or-nothing: it either logs all bodies or none.
+
+Body logging exclusions give operators granular control: suppress request bodies for high-volume operations (like bulk ingestion) while still capturing bodies for searches, index creation, and admin operations that matter for investigation.
+
+##### How it works
+
+Each audit event's action name and REST path are matched against a set of exclusion patterns. When a match is found, the request body field is omitted from the audit event---all other fields (user, IP address, indices, timestamp, etc.) are preserved.
```
DarshitChanpura, line 333: The `:` / `/` / `*` split reads like it drives matching, but in the code it's only used for the warning heuristic — matching is a single combined `WildcardMatcher` tested per layer. Outcome you describe is right, just the framing. Non-blocking.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,308 @@
+---
+layout: default
+title: Standalone audit logging
+parent: Audit logs
+nav_order: 130
```
DarshitChanpura, line 5: Collides with `field-reference.md`, which is also `130`. Use `128`, or `133` to land between field-reference and storage-types (`135`).

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,308 @@
...
+
+Identity information varies by security mode:
+
+- **SSL-only with mTLS**: The client certificate's subject DN is logged as `audit_request_effective_user` (e.g., `CN=my-app,OU=engineering,O=myorg`).
+- **SSL-only without mTLS**: Only the source IP address is captured.
+- **Security disabled**: Only the source IP address is captured.
+
+## Configuration
+
+All standalone audit settings are configured in `opensearch.yml` for initial values and can be dynamically updated at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/). No security index is required.
```
DarshitChanpura, line 109: The `PUT` write path works today; the read-back (and the panel reading state back) is #6392. Same hold — don't merge before #6392 is in.

DarshitChanpura (changes requested):
Blocker: this documents the read-back of standalone audit config via `GET _cluster/settings`, but that behavior comes from opensearch-project/security#6392, which isn't merged yet. In current `main`, `getSettingsFilter()` strips the whole `plugins.security.audit.*` subtree from settings responses in SSL-only mode, so the read path here doesn't work. Hold this until #6392 is merged and released, and pin it to the same version.

Otherwise the settings, defaults, tracked events, body-logging exclusions, MDC keys, and compliance behavior all match the merged code. Couple of inline fixes below.

kolchfa-aws:
@Taiwo435 Would you be able to make the requested changes so we can move forward with this PR?

Taiwo435:
@kolchfa-aws I apologize for the delay. I’ve addressed the review comments and pushed the changes.

kolchfa-aws:
@DarshitChanpura Could you take a look at the changes when you get a chance?

DarshitChanpura (commented):
Re-review of the standalone audit logging docs. I verified each claim against `opensearch-project/security` `main` (`SecuritySettings`, `ConfigConstants`, `OpenSearchSecurityPlugin`, `AuditConfig`, `ComplianceConfig`, `AuditActionFilter`, `AbstractAuditLog`, `AuditMessage`, `Log4JSink`, `InternalOpenSearchSink`) and core `Setting.java`.

Nice work on the coverage here — the new page is thorough and the setting names, static/dynamic split for the new settings, MDC keys, body-exclusion semantics, and bulk all-or-nothing behavior all check out. Most of what follows is a mismatch between documented behavior and what the code currently does, which is easy to hit because several of these paths fail silently.

**Four higher-severity items** (documented procedures that do not work, with no error surfaced):
1. `Setting.Property.Sensitive` is described backwards — it gates dynamic *writes*, it does not redact values (`security-settings.md:248`, `index.md:276`).
2. The audit filter/compliance settings are **not** runtime-dynamic in FGAC mode; only `body_logging_exclusions` is. Updates return `200 OK` and do nothing (`index.md:276`).
3. Neither documented way to enable/disable audit logging works as described — the yml form is ignored in all modes (`index.md:278`).
4. The compliance `opensearch.yml` examples read from legacy `opendistro_security.compliance.history.*` keys, so the `plugins.*` form shown produces zero compliance events (`standalone.md:216`, `229`, `307`).

The rest are accuracy fixes: five settings listed as dynamic have no update consumer, the disabled-categories defaults are incomplete (and the setting replaces rather than merges), `REQUEST_AUDIT` does not carry the REST method/path fields, a REST-vs-TRANSPORT layer contradiction appears on both pages, and the body-exclusion warning condition is `startsWith("/")` rather than `contains("/")`.

Two smaller notes not inline, since they fall outside the diff hunks:
- `security-settings.md:192` still documents `config.index` default as `auditlog6`, which now contradicts the corrected `security-auditlog-YYYY.MM.dd` at `index.md:465` (`InternalOpenSearchSink` confirms `'security-auditlog-'YYYY.MM.dd`). Worth fixing while the PR is open.
- The new page uses the leading `{: .note }` / `{: .warning }` callout form; every other `_security` page puts the marker on the line *after* the paragraph. Only the trailing form renders as a callout.

Happy to help with the wording on any of these.

_install-and-configure/configuring-opensearch/security-settings.md
```diff
@@ -237,6 +245,8 @@ The Security plugin supports the following audit log settings:
 
 - `opendistro_security.audit.config.disabled_transport_categories` (Dynamic): A list of transport layer categories to be ignored by the logger. Valid values are `AUTHENTICATED` and `GRANTED_PRIVILEGES`.
 
+The dynamic audit filter settings (except `plugins.security.audit.config.body_logging_exclusions`) and all `plugins.security.audit.compliance.*` settings are registered with `Setting.Property.Sensitive`, which keeps their values out of diagnostics and logs. In standalone SSL-only mode, any caller can read the non-secret dynamic configuration under `plugins.security.audit.config.*` and `plugins.security.audit.compliance.*` using `GET _cluster/settings`. Credential-bearing sink settings, including endpoint and route credentials, PEM keys, and salt, remain hidden. In FGAC mode, the entire `plugins.security.audit.*` subtree is filtered from settings responses for all callers; this filtering is not role-based. The static `plugins.security.audit.config.action_groups.<NAME>` setting is also not marked sensitive.
```
DarshitChanpura: `Setting.Property.Sensitive` is described backwards here. In core (`Setting.java`, `isSensitive()`), `Sensitive` means **updating the setting dynamically requires security-admin privileges** — it is enforced by `SecurityFilter.handleBlockedSensitiveSettingsUpdate`, which rejects a `ClusterUpdateSettings` request touching a sensitive key unless the caller holds a role in `plugins.security.restapi.roles_enabled`. It does **not** keep values out of diagnostics/logs — that is `Property.Filtered`.

Two consequences worth stating explicitly:
- A reader concludes audit values are redacted. They are not.
- Because `body_logging_exclusions` is *not* sensitive, any caller permitted to update cluster settings can change it, while the others need the elevated role.

Also: this write restriction depends on `SecurityFilter`, which is not installed in SSL-only or security-disabled mode — i.e. it is unenforced in exactly the standalone modes this PR documents.

Separately, the read-visibility sentence omits **security-disabled** mode: `OpenSearchSecurityPlugin.getSettingsFilter()` returns an *empty* list when `plugins.security.disabled: true`, so nothing under `plugins.security.audit.*` is filtered there — including the `endpoints.*`/`routes.*` group settings that hold sink `username`/`password`, `webhook.url`, and PEM content. The claim that those "remain hidden" holds for SSL-only only.

_security/audit-logs/index.md
```diff
@@ -260,11 +273,30 @@ config:
...
...e stored in the `opensearch.yml` file. The audit filter and compliance settings---for example, `log_request_body`, `resolve_indices`, `disabled_categories`, and the `plugins.security.audit.compliance.*` settings---are dynamic: they can be changed at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/) without a node restart. Most of these dynamic settings are registered with `Setting.Property.Sensitive`, which keeps their values out of diagnostics and logs; `body_logging_exclusions` is a dynamic setting that is not marked sensitive. In standalone SSL-only mode, any caller can read the non-secret dynamic audit configuration (`plugins.security.audit.config.*` and `plugins.security.audit.compliance.*`) using `GET _cluster/settings`. Credential-bearing sink settings remain hidden. In FGAC mode, the entire `plugins.security.audit.*` subtree is filtered from settings responses for all callers. This read filtering is not role-based. Other audit settings---such as `action_groups.<NAME>`, `log4j.enable_mdc_routing`, `config.index`, the thread pool settings, and the sink connection settings---are static and require a node restart.
```
DarshitChanpura: Two problems in this paragraph.

**1. The dynamic claim is false on this (standard/FGAC) page.** Every cluster-settings update consumer for `log_request_body`, `resolve_indices`, `disabled_categories`, `enable_rest`/`enable_transport`, `ignore_users`/`ignore_requests`, and `plugins.security.audit.compliance.*` is registered inside `OpenSearchSecurityPlugin.initStandaloneAuditIfEnabled(...)`, which only runs when `enable_standalone: true` (SSL-only or `disabled`). In the FGAC branch only `body_logging_exclusions` gets a consumer.

Failure mode: an FGAC user runs
```json
PUT _cluster/settings
{"persistent":{"plugins.security.audit.config.log_request_body":false}}
```
gets `200 OK`, and bodies keep being logged — FGAC config still comes from `audit.yml`. Worth calling out as a warning, since a silent no-op on an audit control is easy to mistake for working.

**2. `Setting.Property.Sensitive` is described backwards** — same issue as `security-settings.md:248`; it gates dynamic *writes* (requires a `restapi.roles_enabled` role), it does not redact values.

_security/audit-logs/index.md
```diff
@@ -260,11 +273,30 @@ config:
...
+
+#### Enable or disable audit logging
```
DarshitChanpura: Neither route documented in this section works as described.

- **The yml form is silently ignored in all modes.** `AuditConfig.from(Settings)` hardcodes `new AuditConfig(true, ...)` and `SecuritySettings.AUDIT_ENABLED_SETTING.get(settings)` is never called, so `plugins.security.audit.enabled: false` in `opensearch.yml` has no effect.
- **The `PUT _cluster/settings` form only works in standalone mode** (consumer registered at `OpenSearchSecurityPlugin` line ~409).

Since this section sits under "Settings in `opensearch.yml`" on the FGAC page, both halves are misleading here. Suggest removing the section from this page and pointing at the standalone doc for the runtime toggle.

_security/audit-logs/index.md
```diff
@@ -59,7 +64,15 @@ Event | Logged on REST | Logged on transport | Description
 `BAD_HEADERS` | Yes | Yes | An attempt was made to spoof a request to OpenSearch with the Security plugin internal headers.
 `CLUSTER_SETTINGS_CHANGED` | No | Yes | A persistent or transient cluster setting was changed. Disabled by default.
 `INDEX_SETTINGS_CHANGED` | No | Yes | An index setting was changed. Disabled by default.
+`REQUEST_AUDIT` | Yes | No | A REST-layer request was received and processed. Generated in [standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) mode only.
```
DarshitChanpura: This row says `REQUEST_AUDIT` is REST = Yes / transport = No, but the note four lines below says its layer is `TRANSPORT`. The two read as contradictory — worth renaming the column concept (origin vs. layer) or reconciling the row so a reader isn't left guessing which setting suppresses it.

_security/audit-logs/index.md
```diff
@@ -59,7 +64,15 @@ Event | Logged on REST | Logged on transport | Description
...
+`TRANSPORT_AUDIT` | No | Yes | A transport-layer request was received on a node. Generated in [standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) mode only.
+`RESOURCE_ACCESS_GRANTED` | No | Yes | Access to a shared resource was granted. Disabled by default.
+`RESOURCE_ACCESS_DENIED` | No | Yes | Access to a shared resource was denied. Disabled by default.
+`RESOURCE_SHARING_CHANGED` | No | Yes | A resource sharing configuration was changed. Disabled by default.
+
 
+Although `REQUEST_AUDIT` records a REST-origin request (`audit_request_origin: REST`), its internal request layer is `TRANSPORT` (`audit_request_layer: TRANSPORT`). It can therefore be suppressed by `disabled_transport_categories` as well as the unified `disabled_categories` setting.
```
DarshitChanpura: The suppression list is incomplete. `AbstractAuditLog.logRequestAudit()` drops the event if the category appears in `disabled_categories` **or** `disabled_transport_categories` **or** `disabled_rest_categories` (and `AuditActionFilter` short-circuits on the first two).

So "suppressed by `disabled_transport_categories` as well as the unified setting" omits that `disabled_rest_categories` also suppresses it. Someone who adds `REQUEST_AUDIT` to `disabled_rest_categories` expecting the event to survive will lose all of them.

_security/audit-logs/index.md
```diff
@@ -288,12 +320,149 @@ plugins.security.audit.config.disabled_transport_categories:
...
+}
+```
+{% include copy.html %}
+
+Each entry in the list is processed as follows:
+
+1. If the entry matches a defined action group name, that group's patterns are expanded.
+2. If it does not match any group name, the entry is treated as a raw pattern (literal or wildcard).
+
+For entries that do not match a defined group name, a warning is logged if the entry contains none of `:`, `/`, or `*`, since it is unlikely to match an action or REST path. This check only determines whether to log a warning; the entry is still included in the combined matcher.
```
DarshitChanpura: The condition is inverted/incomplete. `rebuildBodyExclusionMatcher()` tests:

```java
!entry.contains(":") && !entry.startsWith("/") && !entry.contains("*")
```

Note `startsWith("/")`, not `contains("/")`. So an entry like `_bulk/items` contains a `/` but does not *start* with one, and therefore **does** log the warning — contrary to "contains none of `:`, `/`, or `*`".

_security/audit-logs/index.md
```diff
@@ -39,9 +46,7 @@ Audit logging is disabled by default. To enable audit logging:
 
 2. Restart each node.
 
-After this initial setup, you can use OpenSearch Dashboards to manage your audit log categories and other settings. In OpenSearch Dashboards, select **Security** and then **Audit logs**. 
-
-An alternative is to specify initial settings for audit logging in the `audit.yml` and `opensearch.yml` files (which file depends on the setting---see [Audit log settings](#audit-log-settings)). Thereafter, you can use Dashboards or the [Audit logs]({{site.url}}{{site.baseurl}}/security/access-control/api/#audit-logs) API to manage and update settings.
+After this initial setup, you can specify settings for audit logging in the `audit.yml` and `opensearch.yml` files (which file depends on the setting---see [Audit log settings](#audit-log-settings)). You can also use the [Audit logs]({{site.url}}{{site.baseurl}}/security/access-control/api/#audit-logs) API to manage and update settings.
```
DarshitChanpura: This drops the OpenSearch Dashboards **Security → Audit logs** path, which still works, and replaces it with the `plugins.security.audit.enabled` route that does not (see my comment at line 278). Net effect: the page is left with no working non-yml method for managing audit settings in FGAC mode.

Same deletion happens again in "Disabling audit logs" (~line 500), where the Dashboards **Enable audit logging** checkbox is replaced by the non-functional `audit.enabled` cluster setting. Suggest restoring the Dashboards path in both places and scoping the `enabled` toggle to the standalone page.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+{% include copy.html %}
+
+Both settings are required:
+- `plugins.security.audit.enable_standalone: true` activates the standalone audit subsystem.
+- `plugins.security.audit.type: <sink>` specifies the audit sink (where events are stored).
+
+After adding these settings, restart each node to activate standalone audit logging.
+
+{: .warning }
+Authentication-only audit categories (`FAILED_LOGIN`, `AUTHENTICATED`, `GRANTED_PRIVILEGES`, `MISSING_PRIVILEGES`) will never produce events in standalone mode because no authentication or authorization decisions occur. A warning is logged at startup listing these inactive categories.
```
DarshitChanpura: Two corrections.

**The list is incomplete.** `AuditCategory.AUTH_ONLY_CATEGORIES` also contains `OPENDISTRO_SECURITY_INDEX_ATTEMPT`, `API_TOKEN_WRITE`, `RESOURCE_ACCESS_GRANTED`, `RESOURCE_ACCESS_DENIED`, and `RESOURCE_SHARING_CHANGED`.

**The startup-warning claim is off.** `warnIfAuthCategoriesEnabled()` subtracts the disabled categories first, so with default settings the logged warning will *not* mention `AUTHENTICATED` or `GRANTED_PRIVILEGES` — two of the four this sentence says it lists.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Supported audit sinks
+
+Standalone audit logging supports the same sinks as standard mode:
+
+Sink type | Description
+:--- | :---
+`internal_opensearch` | Writes audit events to an index on the current OpenSearch cluster.
```
DarshitChanpura, line 57: `debug` is missing from this table (`plugins.security.audit.type: debug`, prints events to stdout). Worth adding with a note that it is for development/troubleshooting only, since the sentence above claims "the same sinks as standard mode".

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+
+For sink-specific configuration options, see [Audit log storage types]({{site.url}}{{site.baseurl}}/security/audit-logs/storage-types/).
+
+## Tracked events
+
+Standalone audit logging introduces two request-tracking categories designed for environments without authentication:
+
+Category | Layer | Description
+:--- | :--- | :---
+`REQUEST_AUDIT` | REST | Captures all REST-layer requests including source IP, target indexes, request body, HTTP headers, and request method. This is the primary event for standalone mode.
```
DarshitChanpura: Two issues on this row.

**Layer contradiction:** the column says Layer = `REST`, but the note three lines down says the internal layer is `TRANSPORT`. Consider renaming the column to "Origin" so the row and the note can both be correct.

**`request method` is not captured.** `audit_rest_request_method` / `audit_rest_request_path` are only set by `AuditMessage.addRestRequestInfo()` (`addPath`/`addRestMethod`), which `AuditActionFilter` never calls — see my comment at line 93.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+`TRANSPORT_AUDIT` | Transport | Captures transport-layer requests between nodes, including shard-level operations (`bulk[s][p]`, `search[phase/query]`), replica writes, and forwarded requests.
+
+Although `REQUEST_AUDIT` originates from REST, its internal request layer is `TRANSPORT`. It can be suppressed by `disabled_transport_categories` as well as the unified `disabled_categories` setting.
```
DarshitChanpura: Same omission as `index.md:74`: `AbstractAuditLog.logRequestAudit()` also drops the event when the category is in `disabled_rest_categories`, so all three settings suppress `REQUEST_AUDIT`, not just `disabled_transport_categories` and the unified setting.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+These categories do not imply any authentication or authorization semantics---they simply record that a request was received and processed.
+
+In addition to these request-tracking categories, standalone audit logging emits the standard document-level compliance categories when [compliance tracking](#compliance-tracking) is enabled:
+
+Category | Description
+:--- | :---
+`COMPLIANCE_DOC_WRITE` | A document was written to a watched index. See [Document write tracking](#document-write-tracking).
+`COMPLIANCE_DOC_READ` | A watched field was read from a watched index. See [Document read tracking](#document-read-tracking).
+
+Compliance events are governed only by the compliance settings (watched indexes/fields and `compliance.enabled`). They are not affected by `disabled_categories`, which applies only to `REQUEST_AUDIT` and `TRANSPORT_AUDIT`.
```
DarshitChanpura: "`disabled_categories` ... applies only to `REQUEST_AUDIT` and `TRANSPORT_AUDIT`" overstates the scope — it applies to **every** REST/transport category. (The compliance carve-out itself is correct: `checkComplianceFilter` bypasses it.)

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Event fields
+
+Each `REQUEST_AUDIT` event includes:
+
+- `@timestamp` --- When the event occurred
+- `audit_cluster_name`, `audit_node_name`, `audit_node_id` --- Cluster and node identity
+- `audit_rest_request_method`, `audit_rest_request_path` --- HTTP method and path
```
DarshitChanpura: `REQUEST_AUDIT` events do not carry these two fields. `audit_rest_request_method` / `audit_rest_request_path` are only populated by `AuditMessage.addRestRequestInfo()`, which `AuditActionFilter` never calls — it sets remote address, privilege/action, request type, indices, task id, effective user, filtered REST headers, and body only.

Anyone building a dashboard on `audit_rest_request_path` from standalone events gets nothing. Suggest replacing this bullet with `audit_request_privilege` (the transport action) and noting that the REST method/path fields are FGAC-only.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Configuration
+
+Configure initial standalone audit settings in `opensearch.yml`. Dynamic settings can be updated at runtime using the [Cluster settings API]({{site.url}}{{site.baseurl}}/api-reference/cluster-api/cluster-settings/). Static settings, including `enable_standalone`, `action_groups.<NAME>`, and sink connection settings, require a node restart. No security index is required.
+
+Standalone mode does not use the [Audit logs]({{site.url}}{{site.baseurl}}/security/access-control/api/#audit-logs) REST API or `audit.yml`---those manage the security index and require fine-grained access control. In standalone mode, use the Cluster settings API instead.
+
+### Dynamic configuration
+
+All filter and compliance settings can be changed at runtime without restarting the cluster:
```
DarshitChanpura: "All filter and compliance settings" is not accurate — five of them have no `addSettingsUpdateConsumer` and appear only in `getSettings()`:

- `plugins.security.audit.config.ignore_headers`
- `plugins.security.audit.compliance.external_config`
- `plugins.security.audit.compliance.internal_config`
- `plugins.security.audit.compliance.read_ignore_users`
- `plugins.security.audit.compliance.write_ignore_users`

A `PUT _cluster/settings` on any of these is accepted and has no effect, before or after restart. They are also all listed in the "Dynamic settings reference" below and under "All compliance settings are dynamic" (line 253).

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+  }
+}
+```
+{% include copy.html %}
+
+Dynamic settings override the values in `opensearch.yml` and persist across cluster restarts.
+
+### Dynamic settings reference
+
+The following settings are all dynamic---they can be set in `opensearch.yml` for initial values and updated at runtime via `PUT _cluster/settings`. Types are shown as placeholders:
```
DarshitChanpura: Two things this reference block gets wrong.

**`plugins.security.audit.enabled` is runtime-only.** `AuditConfig.from(Settings)` hardcodes `true` and never reads `AUDIT_ENABLED_SETTING`, so setting it in `opensearch.yml` has no effect in any mode. Only `PUT _cluster/settings` works (standalone only).

**The `plugins.security.audit.compliance.*` keys do not work from `opensearch.yml`.** `ComplianceConfig.from(Settings)` reads everything except `enabled` from the **legacy** keys — `opendistro_security.compliance.history.write.watched_indices`, `...read.watched_fields`, `...write.log_diffs`, `...{read,write}.metadata_only`, `...{read,write}.ignore_users`, `...external_config_enabled`, `...internal_config_enabled`. The `plugins.*` form is only *registered* (so the node starts cleanly) and is honored solely through the runtime update consumers.

So "can be set in `opensearch.yml` for initial values" is false for the compliance block. See my comment at line 216.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+Setting | Default | Description
+:--- | :--- | :---
+`plugins.security.audit.config.enable_rest` | `true` | Enable REST-layer audit events.
+`plugins.security.audit.config.enable_transport` | `true` | Enable transport-layer audit events.
+`plugins.security.audit.config.log_request_body` | `true` | Include the request body in audit events.
+`plugins.security.audit.config.resolve_indices` | `true` | Resolve wildcard index patterns to concrete indexes.
+`plugins.security.audit.config.resolve_bulk_requests` | `false` | Log individual sub-operations in bulk requests.
+`plugins.security.audit.config.exclude_sensitive_headers` | `true` | Exclude sensitive headers (e.g., `Authorization`) from audit events.
+`plugins.security.audit.config.disabled_categories` | `[]` | Request-tracking categories to disable (e.g., `["REQUEST_AUDIT"]`). Does not affect `COMPLIANCE_*` categories.
+`plugins.security.audit.config.disabled_rest_categories` | `["AUTHENTICATED", "GRANTED_PRIVILEGES"]` | REST-layer categories to disable. On a deprecation path---prefer `disabled_categories`.
```
DarshitChanpura: These documented defaults are incomplete. The effective defaults come from `ConfigConstants.OPENDISTRO_SECURITY_AUDIT_DISABLED_*_CATEGORIES_DEFAULT`, which is what `AuditConfig.Filter.from()` uses:

- REST: `AUTHENTICATED`, `GRANTED_PRIVILEGES`, `RESOURCE_ACCESS_GRANTED`, `RESOURCE_ACCESS_DENIED`, `RESOURCE_SHARING_CHANGED`
- Transport: those five plus `CLUSTER_SETTINGS_CHANGED`, `INDEX_SETTINGS_CHANGED`

`["AUTHENTICATED", "GRANTED_PRIVILEGES"]` also contradicts this PR's own `index.md` table, which marks the three `RESOURCE_*` categories and the two `*_SETTINGS_CHANGED` categories "Disabled by default".

This matters because the setting replaces the list rather than merging: a user who copies the documented "default" silently re-enables five categories. Worth a warning to that effect.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+
+Document-level compliance tracking works in standalone mode for both reads and writes.
+
+### Document write tracking
+
+To track writes to specific indexes, configure the watched indexes:
+
+```yml
+plugins.security.audit.compliance.enabled: true
+plugins.security.audit.compliance.write_watched_indices:
```
DarshitChanpura: This example does not work. `ComplianceConfig.from(Settings)` reads watched indices from the legacy key `opendistro_security.compliance.history.write.watched_indices`; the `plugins.security.audit.compliance.write_watched_indices` form in `opensearch.yml` is accepted without error but never read at startup.

Failure mode: a user copies this block into `opensearch.yml`, restarts, and gets **zero** `COMPLIANCE_DOC_WRITE` events with no error anywhere.

Suggest showing the runtime `PUT _cluster/settings` form as the primary example (which does work) and the legacy `opendistro_security.compliance.history.*` keys for startup configuration. Same applies to the read example at ~229 and the security-disabled example at ~307.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+`plugins.security.audit.compliance.read_metadata_only` | `false` | Log only metadata for read events (no field values).
+`plugins.security.audit.compliance.write_log_diffs` | `false` | Include diffs between old and new document content.
+`plugins.security.audit.compliance.write_watched_indices` | `[]` | Index patterns to watch for write compliance events.
+`plugins.security.audit.compliance.read_watched_fields` | `[]` | Index-and-fields patterns to watch for read compliance events. Each entry is a comma-separated string: `<index-pattern>,<field-pattern>,...`.
+`plugins.security.audit.compliance.write_ignore_users` | `["kibanaserver"]` | Users whose document writes are not tracked for compliance.
+`plugins.security.audit.compliance.read_ignore_users` | `["kibanaserver"]` | Users whose document reads are not tracked for compliance.
+`plugins.security.audit.compliance.external_config` | `false` | Log the external configuration (`opensearch.yml` and environment) once at startup.
+`plugins.security.audit.compliance.internal_config` | `false` | Log changes to the internal security configuration.
+
+All compliance settings are dynamic and can be updated via `PUT _cluster/settings`.
```
DarshitChanpura: Not accurate for four of the settings in the table above — `write_ignore_users`, `read_ignore_users`, `external_config`, and `internal_config` have no runtime update consumer (see my comment at line 118). They must be set through their legacy `opendistro_security.*` keys plus a node restart.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,313 @@
...
+plugins.security.audit.type: internal_opensearch
+
+# Audit filter settings
+plugins.security.audit.config.log_request_body: true
+plugins.security.audit.config.resolve_indices: true
+plugins.security.audit.config.resolve_bulk_requests: true
+
+# Compliance tracking
+plugins.security.audit.compliance.enabled: true
+plugins.security.audit.compliance.write_watched_indices:
```
DarshitChanpura: Same issue as line 216: `plugins.security.audit.compliance.write_watched_indices` is not read from `opensearch.yml`. As written, this example config produces no compliance events. Use `opendistro_security.compliance.history.write.watched_indices` here, or move the compliance part to a runtime `PUT _cluster/settings` call.

kolchfa-aws:
@Taiwo435 Please see the latest review and address the comments when you can. Also, please resolve merge conflicts. Thanks!

DarshitChanpura:
Heads-up on an overlap: #13123 touches this same section of `_security/audit-logs/index.md` and is worth rebasing onto rather than merging around.

It adds a note right after "2. Restart each node." explaining that the shipped `audit.yml` has `config.enabled: true` yet audit logging still does nothing without a storage type in `opensearch.yml`. I verified its claims against `main` and they hold: `AuditLogImpl:89` computes `enabled = auditConfig.isEnabled() && messageRouterEnabled`, `AuditMessageRouter:72` returns `defaultSink != null`, and `SinkProvider:85` logs `Default endpoint could not be created…` when `plugins.security.audit.type` is unset.

Two implications for this PR:

**1. It confirms the `#### Enable or disable audit logging` section should be dropped, not reworded.** My earlier comment at `index.md:278` noted that `plugins.security.audit.enabled` in `opensearch.yml` is never read — `AuditConfig.from(Settings)` hardcodes `true` (`AuditConfig.java:127`) and the only consumer is the runtime one at `OpenSearchSecurityPlugin:409` → `setEnabled()`, registered solely under `initStandaloneAuditIfEnabled(...)`.

These are two different switches, so #13123 and this PR are not in conflict on the code — but they are on the page:

| Switch | Read from | Honored at startup? |
| --- | --- | --- |
| `config.enabled` in `audit.yml` | Jackson `@JsonProperty("enabled")` → `isEnabled()` | Yes, and gated on a storage type |
| `plugins.security.audit.enabled` node setting | `AuditConfig.from(Settings)` — hardcoded `true` | No; runtime-only, standalone mode only |

If both land as-is, the page ends up describing three ways to enable audit logging, one of which has no effect in FGAC. #13123's note is the correct account for FGAC; the toggle in this PR is only meaningful at runtime in standalone mode, so it belongs on the standalone page.

**2. This branch is stale on `main` independent of that.** #13123's diff context still shows the original "you can use OpenSearch Dashboards" paragraph, and it links the Audit logs API as `/security/api/audit/` where this branch uses `/security/access-control/api/#audit-logs`. So a rebase is needed either way, and whichever of the two merges second will conflict in this hunk.

Suggest rebasing after #13123 merges and keeping its note as the single enable/disable explanation for FGAC.

Taiwo435:
Thanks for flagging this and for verifying the source claims, good to know #13123 already got merged.

Once I rebase, the `#### Enable or disable audit logging` section here will point to the standalone page instead of describing the plugins.security.audit.enabled toggle, which should line up with keeping #13123's note as the single FGAC-facing explanation.

kolchfa-aws:
@DarshitChanpura Could you take a look at this PR again when you get a chance?

_install-and-configure/configuring-opensearch/security-settings.md
```diff
@@ -195,7 +195,7 @@ The Security plugin supports the following audit log settings:
 
 - `plugins.security.audit.config.http_endpoints` (Static): A list of endpoints for `localhost`.
 
-- `plugins.security.audit.config.index` (Static): The audit log index. The default is `auditlog6`. The index can be static or an index that includes a date so that it rotates on a daily basis, for example, `"'auditlog6-'YYYY.MM.dd"`. In either case, make sure to secure the index properly.
+- `plugins.security.audit.config.index` (Static): The audit log index. The default is `security-auditlog`. The index can be static or an index that includes a date so that it rotates on a daily basis, for example, `"'security-auditlog-'YYYY.MM.dd"`. In either case, make sure to secure the index properly.
```
DarshitChanpura, line 198: The default is the date-rolling pattern, not a static index: `"'security-auditlog-'YYYY.MM.dd"`. `InternalOpenSearchSinkIntegrationTestConcreteIndex` documents it as "the default pattern `'security-auditlog-'YYYY.MM.dd` produces daily" indexes.

As written this also disagrees with `_security/audit-logs/index.md`, which now correctly says audit events go to a daily rolling index named `security-auditlog-YYYY.MM.dd`.

_security/audit-logs/standalone.md
```diff
@@ -0,0 +1,317 @@
...
+`plugins.security.audit.config.disabled_categories` | `[]` | Request-tracking categories to disable (e.g., `["REQUEST_AUDIT"]`). Does not affect `COMPLIANCE_*` categories.
+`plugins.security.audit.config.disabled_rest_categories` | `["AUTHENTICATED", "GRANTED_PRIVILEGES", "RESOURCE_ACCESS_GRANTED", "RESOURCE_ACCESS_DENIED", "RESOURCE_SHARING_CHANGED"]` | REST-layer categories to disable. On a deprecation path---prefer `disabled_categories`.
+`plugins.security.audit.config.disabled_transport_categories` | `["AUTHENTICATED", "GRANTED_PRIVILEGES", "RESOURCE_ACCESS_GRANTED", "RESOURCE_ACCESS_DENIED", "RESOURCE_SHARING_CHANGED", "CLUSTER_SETTINGS_CHANGED", "INDEX_SETTINGS_CHANGED"]` | Transport-layer categories to disable. On a deprecation path---prefer `disabled_categories`.
+`plugins.security.audit.config.ignore_users` | `["kibanaserver"]` | Users whose requests are not logged.
+`plugins.security.audit.config.ignore_requests` | `[]` | Action patterns or REST paths to exclude (e.g., `["cluster:monitor/*"]`).
+`plugins.security.audit.config.ignore_headers` | `[]` | HTTP headers to exclude from audit events.
```
DarshitChanpura, line 189: `ignore_headers` is one of the five settings with no registered update consumer, per the list under "Dynamic configuration" earlier on this page. This row does not carry that qualifier, and the setting also still appears in the dynamic settings reference block above, so a reader who lands on either spot alone will read it as runtime-updatable.

Suggest adding the same "Not updatable at runtime" qualifier the compliance table uses, or dropping it from the yml reference block.

_security/audit-logs/index.md
```diff
@@ -62,7 +67,15 @@ Event | Logged on REST | Logged on transport | Description
 `BAD_HEADERS` | Yes | Yes | An attempt was made to spoof a request to OpenSearch with the Security plugin internal headers.
 `CLUSTER_SETTINGS_CHANGED` | No | Yes | A persistent or transient cluster setting was changed. Disabled by default.
 `INDEX_SETTINGS_CHANGED` | No | Yes | An index setting was changed. Disabled by default.
+`REQUEST_AUDIT` | Yes | Yes | A REST request was received and processed. Generated in [standalone audit logging]({{site.url}}{{site.baseurl}}/security/audit-logs/standalone/) mode only. See the note below the table for how this event is suppressed.
```
DarshitChanpura, line 70: `REQUEST_AUDIT` fires once, from `AuditActionFilter`, and is recorded with `audit_request_layer: TRANSPORT`. Marking both columns `Yes` reads as two events on two layers. The Origin column used on the standalone page handles this case more clearly.

The note below the table does disambiguate, so this is cosmetic.

DarshitChanpura (commented):
All 12 items from the previous review are addressed, and the rebase onto the merged PR 13123 resolved the duplicate enable-path problem cleanly.

Re-verified against security `main`: the `Property.Sensitive` write-authorization semantics, the FGAC-vs-standalone update-consumer split, the legacy `opendistro_security.compliance.history.*` startup keys, the five settings with no update consumer, the corrected disabled-category defaults plus the replaces-rather-than-merges warning, and `audit_request_privilege` replacing the REST method/path fields. Callouts are all trailing markers now and there is no duplicated prose from the rebase.

Three minor things below, none blocking.

kolchfa-aws:
@Taiwo435 A couple of more tech review suggestions before we can move on to doc review. Could you address when you get a chance? Thanks!

Taiwo435:
@DarshitChanpura and @kolchfa-aws Thanks for the re-review. Fixed the two accuracy items: config.index's default is now correctly described as the date-rolling pattern ('security-auditlog-'YYYY.MM.dd), and ignore_headers now carries the same "no registered update consumer" note where it's listed in the filter settings section, matching the compliance table's pattern.
Left the REQUEST_AUDIT Yes/Yes row as-is on this page, since you flagged it as cosmetic and the note below it already disambiguates.

DarshitChanpura (approved):
Both substantive items from the last round are fixed: `config.index` now documents the date-rolling default, and `ignore_headers` carries the no-update-consumer qualifier next to its table row.

Re-checked the twelve findings from the first round at this head; all still hold, and the commit touched only the two flagged lines. Leaving the `REQUEST_AUDIT | Yes | Yes` column as-is is fine, the note below the table covers it.

LGTM.

kolchfa-aws (approved):
Thank you, @Taiwo435! Doc review complete and changes applied.

PR #12928, opened 2026-08-18
Expand system index documentation

cwperks:
## Description

Expand the system index documentation to cover:

- descriptor-based registration, automatic creation, and system thread pools
- the exact system index patterns registered by the standard distribution
- Security plugin write protection, scrubbed reads, explicit permissions, and audit events
- guidance for plugin developers using `SystemIndexPlugin` and plugin identity

The inventory was verified by instrumenting `SystemIndexRegistry` in a throwaway matching Core build and starting the OpenSearch 3.7.0 standard distribution. This enumerated all descriptors at startup, including indexes that had not yet been created. The diagnostic instrumentation is not part of this change.

## Testing

- `git diff --check`
- Compared the documentation table programmatically with the startup registry dump: 66 unique patterns in each, with no differences

The full Jekyll build could not run locally because the available system Ruby is 2.6.10 and the current documentation dependencies require Ruby 3.1 or later.


kolchfa-aws (approved):
Thank you, @cwperks! Doc review complete and changes applied.

PR #12930, opened 2026-08-18
Fix breadcrumbs in search results and other functionality

### LINKED ISSUE #12497: Search Breadcrumbs collapse everything under OpenSearch X.X version
I am searching for the OS Benchmark tools `--exclude-tasks` and it took me a while to realize that yes, the docs are being indexed, but they are hidden visually in the search results:

Image

I would expect something like "OpenSearch Benchmark 2.0 › Reference › Command reference".

Here is my query: https://docs.opensearch.org/latest/search.html?q=exclude-task

This pattern shows up for some of our other top level projects like DataPrepper:

Image

kolchfa-aws:
Closes #12497 

This PR fixes breadcrumb paths for search results (both in the dropdown and on the search result page). Additionally, it fixes some existing search functionality. Here is the list of changes:

In the indexer (_plugins/search-indexer.rb):
1. Updates how the collections are read since the collections were rearranged. The config moved to `opensearch_collection`, so the lookup returned nil for every document. 
2. Handles the standalone sections (Data Prepper, Benchmark, Clients, and both Migration Assistant sections) as a single collection. Instead of rendering a sample Benchmark page breadcrumb as "OpenSearch 3.8 › Reference › Command reference" (the parent is nil so Benchmark does not show), renders the breadcrumb as "OpenSearch 3.8 › OpenSearch Benchmark › Reference › Command reference". Only the OpenSearch section is versioned, so only its crumb gets a version number now.
3. Includes great_grand_parent (the added navigation level) in breadcrumbs so all 4 levels are included when constructing a benchmark path.
4. Shows the page's H1 as the result instead of the front matter title. Users can now differentiate their search results for "append", for example, because the results will say "Append processor" and "append command" rather than just "append".
5. Version and status labels are not indexed as body text. 420 of production snippets begin with "Introduced 2.8" instead of the first sentence. Label and button text and HTML comments (the  directives), which previously showed up in the search results, are not indexed.
6. The search page and the 404 page are not indexed as results.
7. Serving locally duplicated the whole site in the index on every save. @data was an array in module state that outlives a build, so each rebuild appended another copy; emptying it per build instead truncated the index under --incremental. Keying by URL is correct in both modes. Duplicates in the JSON overwrite the previous ones rather than duplicating.

In the renderer (assets/js/search.js):
1. The version was matched anywhere in the path. On `/latest/migration-assistant/playbook-solr-8.11-to-opensearch-3/`, a search queried an 8.11 index that does not exist and returned nothing. Only the first path segment is read now.
2. Snippets ended mid-word because the API cuts at a fixed length. The partial word is dropped and an ellipsis is added, signifying that this is not a complete sentence.

On the results page (_layouts/search_layout.html):
1. Searching again on the results page updates the URL, so a reload or a copied link replays the query on screen rather than the previous one.
2. Each new query adds a history entry, so pressing the back button traces back through the searches instead of leaving the page. Rerunning the same query (by selecting or clearing a category checkbox) replaces its entry rather than adding one, and the back and forward buttons rerun the query to which they return.
3. Queries having a percent sign (like "100% recall") produce search results instead of a blank page. 

### Checklist

### DIFF
--- a/_layouts/search_layout.html
+++ b/_layouts/search_layout.html
@@ -123,30 +123,65 @@ layout: table_wrappers
         const allCheckbox = document.getElementById('categoryAll');
         return document.querySelectorAll('input[name="categoryGroup"]:checked');
     }
-    
-    function triggerSearch(query) {
+
+    /* The query in the address bar, which is the query this page is showing results for. */
+    function getUrlQuery() {
+        return new URLSearchParams(window.location.search).get('q') || '';
+    }
+
+    /* Shows the results for a query and does nothing else. Separate from `triggerSearch` so that
+       going back to an earlier search can run it again without changing the history. */
+    function runSearch(query) {
       const searchResultsHeader = document.getElementById('searchPageResultsHeader');
+      const resultsContainer = document.getElementById('searchPageResultsContainer');
+
+      if (!query) {
+        searchResultsHeader.textContent = 'Enter a search query.';
+        resultsContainer.innerHTML = '';
+        return;
+      }
+
+      if (getSelectedCategory().length === 0) {
+        searchResultsHeader.textContent = 'Select the result type for your search.';
+        resultsContainer.innerHTML = '';
+        return;
+      }
+
+      const searchType = Array.from(getSelectedCategory()).map(element => element.value).join(',');
+      searchResultsHeader.textContent = `Search results for "${truncateSentence(query, 20)}"`;
+      doResultsPageSearch(query, searchType, getDocsVersion());
+    }
+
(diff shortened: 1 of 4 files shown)


kolchfa-aws:
Screenshot 2026-08-17 at 8 30 20 PM

kolchfa-aws:
Screenshot 2026-08-17 at 8 34 33 PM

kolchfa-aws:
https://github.com/user-attachments/assets/15a9d68f-b3b5-4a17-94cc-7a844456b560

peterzhuamazon (approved):

PR #12955, opened 2026-08-21
Add a dedicated RRF page and additional info to hybrid search docs

kolchfa-aws:
Adds a dedicated RRF page with a worked example. Augments existing hybrid search documentation (including score ranker processor and normalization processor documentation) with information from the related blogs:

- https://opensearch.org/blog/How-does-the-rank-normalization-work-in-hybrid-search/
- https://opensearch.org/blog/introducing-reciprocal-rank-fusion-hybrid-search/
- https://opensearch.org/blog/semantic-science-benchmarks/

### Checklist


_search-plugins/search-pipelines/normalization-processor.md
```diff
@@ -17,6 +18,208 @@ The `normalization-processor` is a search phase results processor that runs betw
...
+Use the following table to compare the three techniques.
+
+| | `min_max` | `l2` | `z_score` |
+| :--- | :--- | :--- | :--- |
+| Output range | [0.0, 1.0], with `0.0` replaced by `0.001` | [0.0, 1.0] | Not a fixed range, with all values less than or equal to `0.0` replaced by `0.001` |
+| Score of the highest-scoring document in a clause | Always `1.0` | Less than `1.0`, unless the clause returns a single document | Varies, and can exceed `1.0` |
+| Score ratios within a clause | Not preserved | Preserved | Not preserved |
+| Statistics used | Lowest and highest scores | All scores | All scores |
+| Influence of a single extreme score | High, because the score defines an endpoint of the range | Lower, because the score is one term of the norm | Lower, because the score shifts the mean and standard deviation |
+| Ordering below the clause mean | Preserved | Preserved | Discarded |
```
martin-gaievski: for `min_max` there is an edge case when order is not preserved.  min_max floors only an exact 0.0f, so any value in (0, 0.001) passes through unfloored. Example scenario: raw [10000, 9, 8] translated to [1.0, 0.00010008, 0.001] after normalization. Raw-9 document normalizes below the raw-8 document, that's rank inversion whenever score spread exceeds ~1000×.

_search-plugins/search-pipelines/normalization-processor.md
```diff
@@ -17,6 +18,208 @@ The `normalization-processor` is a search phase results processor that runs betw
...
+| Supported combination techniques | All | All | `arithmetic_mean` |
+
+The following guidance applies to each technique:
+
+- Use `min_max`, the default, when you want normalized scores on a predictable scale that you can reason about directly, or when you want to normalize against fixed thresholds through the `lower_bounds` and `upper_bounds` parameters.
+- Use `l2` when the ratios between the scores within a query clause carry information that the final ranking should reflect, or when a clause returns occasional extreme scores that would compress the rest of the range under `min_max`.
```
martin-gaievski: this suggests readers to use l2 for a problem it doesn't solve. "Use l2 … when a clause returns occasional extreme scores that would compress the rest of the range under min_max." Because ‖s‖₂ ≥ max(s) > max(s) − min(s), l2's gaps are always ≤ min_max's. [100, 10, 9, 8]: min_max separates the non-outliers by 0.0207, l2 by 0.0198. Reader who follows this switches techniques and gets either nothing or slightly worse.

_search-plugins/search-pipelines/normalization-processor.md
```diff
@@ -17,6 +18,208 @@ The `normalization-processor` is a search phase results processor that runs betw
...
+
+Use the following table to compare the three techniques.
+
+| | `min_max` | `l2` | `z_score` |
+| :--- | :--- | :--- | :--- |
+| Output range | [0.0, 1.0], with `0.0` replaced by `0.001` | [0.0, 1.0] | Not a fixed range, with all values less than or equal to `0.0` replaced by `0.001` |
+| Score of the highest-scoring document in a clause | Always `1.0` | Less than `1.0`, unless the clause returns a single document | Varies, and can exceed `1.0` |
+| Score ratios within a clause | Not preserved | Preserved | Not preserved |
+| Statistics used | Lowest and highest scores | All scores | All scores |
+| Influence of a single extreme score | High, because the score defines an endpoint of the range | Lower, because the score is one term of the norm | Lower, because the score shifts the mean and standard deviation |
```
martin-gaievski: This seems to be revered -  z_score is most fragile. [100, 10, 9, 8] → [1.4998, 0.001, 0.001, 0.001], one outlier erases the ordering of every other document in the clause. min_max keeps all four distinct.

kolchfa-aws: Thank you, @martin-gaievski! All 3 comments addressed in this commit https://github.com/opensearch-project/documentation-website/pull/12955/commits/9c9e1bfce35e87967e47b548c5c170e0bc869cd6

martin-gaievski (approved):
LGTM — all three previous comments addressed correctly. One minor nit:

**normalization-processor.md, z_score section, paragraph 3:**

> "A document whose score is exactly equal to the mean of its clause is a special case: it receives the highest raw score in the clause rather than a z-score."

This reads as a general rule, but it only applies when sd=0 (all scores equal, making division undefined). When sd≠0 and a document happens to equal the mean, its z-score is 0 → replaced by 0.001 per the earlier rule. Consider rewording to clarify this is the all-scores-equal / single-document edge case only.

nit, not blocking approval.

_search-plugins/search-pipelines/normalization-processor.md
```diff
@@ -3,6 +3,7 @@ layout: default
 title: Normalization
 nav_order: 70
 has_children: false
+has_math: true
```
peterzhuamazon, line 6: What is this for? Use LaTex?

peterzhuamazon (approved):

_search-plugins/search-pipelines/normalization-processor.md
kolchfa-aws: Yes, this is for LaTex use by the MathJax plugin so the math formulas render correctly.

PR #12959, opened 2026-08-23
[DOC] Add documentation of client certificate (mutual TLS) authentication for ML connectors

ishag4:
### Description
Connectors can now authenticate to remote endpoints with a client certificate (mutual TLS), in addition to the existing credential-based methods.

### Issues Resolved
Closes #12927 

### Version
3.9

### Frontend features
_If you're submitting documentation for an OpenSearch Dashboards feature, add a video that shows how a user will interact with the UI step by step. A voiceover is optional._ 

### Checklist
- [x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and subject to the [Developers Certificate of Origin](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #12987, opened 2026-08-27
Add Google Cloud Vertex AI connector documentation

tonypiazza:
### Description

Documents the new first-class `google_cloud` connector protocol for Google Cloud Vertex AI, added in ml-commons ([opensearch-project/ml-commons#4921](https://github.com/opensearch-project/ml-commons/pull/4921)). The protocol mints and refreshes GCP OAuth2 access tokens automatically, mirroring the existing `aws_sigv4` strategy, so users no longer supply or rotate an access token by hand.

To keep cloud-provider authentication protocols symmetric and prevent the Connectors page from growing unbounded as more providers are added, connector protocols are documented as peer child pages under **Connectors** rather than as inline sections of a single page. This is the same direction #12959 took, so the pages sit alongside the new **Connector authentication** and **Dynamic header substitution** pages:

- **New page `gcp-connectors.md`** — the `google_cloud` protocol: the `plugins.ml_commons.connector.vertexai_enabled` prerequisite (disabled by default), both authentication modes (service-account key and Application Default Credentials / Workload Identity), and the credential and parameter options.
- **New page `aws-connectors.md`** — the existing Amazon SageMaker and Bedrock (`aws_sigv4`) examples, moved out of `connectors.md` and regrouped so the AWS content is contiguous, with `redirect_from` entries for the old anchors.
- **`connectors.md`** — adds a "Native cloud provider authentication protocols" list linking the child pages, a Google Cloud Vertex AI entry in the platform list, and an updated H1 (see note below).
- **`supported-connectors.md`** — replaces the legacy manual-token Vertex AI row with the four new `google_cloud` blueprints (Gemini, embeddings, streaming, batch).

**Note on the `connectors.md` H1.** This PR changes it from "Creating connectors for third-party ML platforms" to "Connectors for third-party ML platforms". Now that page-creation guidance lives in child pages, the page is a landing page for connector concepts rather than a how-to, so the shorter title matches its scope and its nav label. Flagging it explicitly since it survived the rebase over #12959 and is a deliberate change, not a merge artifact.

### Issues Resolved

N/A

### Version

3.9

### Frontend features

N/A

### Checklist


kolchfa-aws:
@jiapingzeng @mingshl Could you review this PR?

kolchfa-aws:
@tonypiazza Is this going into 3.9?

_ml-commons-plugin/remote-models/gcp-connectors.md
```diff
@@ -0,0 +1,144 @@
...
+
+Use ADC mode only on Google Cloud-hosted nodes. It resolves credentials from the node environment, which includes contacting the Google Cloud metadata server. The service account key and ADC modes are mutually exclusive: in ADC mode, do not include `private_key` or `client_email`.
+
+## Credential and parameter options
+
+The `credential` object contains the following options for service-account key authentication:
+
+- `private_key`: Required in service-account key mode. The service account's private key. Omit in ADC mode.
+- `client_email`: Required in service-account key mode. The service account's client email. Omit in ADC mode.
+- `token_uri`: Optional. The Google OAuth2 token endpoint. Default is `https://oauth2.googleapis.com/token`. When set, it is restricted to Google token endpoints (`*.googleapis.com` over HTTPS).
```
jiapingzeng: From code [here](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleCredentialProvider.java#L88-L90) it seems that `token_uri` host is restricted to `oauth2.googleapis.com`. Should we clarify that here?

tonypiazza: @jiapingzeng  Good catch, and you're right. The check in `validateTokenUri` is an exact host match against `oauth2.googleapis.com`, not a `*.googleapis.com` suffix match, so my wording was both inaccurate and more permissive than the code. Since the guard exists to prevent the node from sending the signed JWT to an arbitrary host, describing it loosely is the wrong tradeoff.

Fixed in 73323d893:

- `token_uri`: Optional. The Google OAuth2 token endpoint. Default is `https://oauth2.googleapis.com/token`. If you set this option, the value must use HTTPS and the host must be exactly `oauth2.googleapis.com`. Any other host is rejected when the connector is created.

tonypiazza:
@kolchfa-aws Yes, 3.9. The ml-commons PR ([opensearch-project/ml-commons#4921](https://github.com/opensearch-project/ml-commons/pull/4921)) is labeled `v3.9.0` and merged to `main` on Aug 26; there's no backport to 3.8. I've updated the Version field in the description and added an **Introduced 3.9** label to the new page in 73323d893.

jiapingzeng (approved):
LGTM!

tonypiazza:
@jiapingzeng Thanks for the approval. Heads up that I've rebased onto `main` to resolve a conflict with #12959, so the approval is now attached to the old commit and GitHub has marked it stale.

One more small commit is coming right behind this (nav_order renumbering, see below), so please hold off re-approving until it lands to save you doing it twice. I'll comment here when it's pushed.

The rebase was mechanical. #12959 split `connectors.md` into **Connector authentication** and **Dynamic header substitution** child pages while this PR moves the `aws_sigv4` examples into **Amazon connectors**, so both sides removed content from the same regions. Both conflict hunks resolved to deletion, with the relocated content preserved on the respective new pages. Two other effects worth noting:

- **The `guardrails.md` fix is dropped from this PR.** #12959 made the identical `has_children: true` -> `false` change, so the diff is now 4 files instead of 5.
- **The child-page structure now matches yours.** `Amazon connectors` and `Google Cloud Vertex AI connector` are renumbered to `nav_order` 50 and 60 to follow the 10-spaced convention, so the six children under **Connectors** run 10/20/30/40/50/60 alongside `Connector authentication` (30) and `Dynamic header substitution` (40).

I verified locally that the site builds, all JSON samples in `connectors.md` still parse, every relocated section survives on its new page, and the six child pages nest correctly in the sidebar. I've also added a note to the description about the `connectors.md` H1 change, which is deliberate rather than a merge artifact.

tonypiazza:
@jiapingzeng The renumbering commit is pushed. Head is now `6ebe718c1` and all checks are green, so this is ready for re-approval whenever you have a moment.

That was the only change: `nav_order` 62 -> 50 on **Amazon connectors** and 63 -> 60 on **Google Cloud Vertex AI connector**, two lines total. The six children under **Connectors** now run 10/20/30/40/50/60.

jiapingzeng (approved):

tonypiazza:
@kolchfa-aws This is ready for doc review when you have time. @jiapingzeng completed the tech review and approved the current head (`6ebe718c1`), and all checks are green.

Two things worth flagging, both from the rebase over #12959:

- **The `guardrails.md` fix is no longer part of this PR.** #12959 made the identical `has_children` change, so the diff is 4 files rather than the 5 my original description mentioned. The description is updated.
- **The `connectors.md` H1** changes from "Creating connectors for third-party ML platforms" to "Connectors for third-party ML platforms". Now that the how-to content lives on child pages, the page reads as a landing page for connector concepts, and the shorter title matches its nav label. It's deliberate rather than a merge artifact.

The child pages under **Connectors** now run 10/20/30/40/50/60, with **Amazon connectors** and **Google Cloud Vertex AI connector** alongside the **Connector authentication** and **Dynamic header substitution** pages from #12959.

_ml-commons-plugin/remote-models/google-cloud.md
```diff
@@ -0,0 +1,150 @@
...
+
+When `protocol` is set to `google_cloud`, the `parameters` object supports the following fields.
+
+| Field | Data type | Required/Optional | Description |
+|:---|:---|:---|:---|
+| `project_id` | String | Required | Your Google Cloud project ID. |
+| `location` | String | Required | The Vertex AI region, for example, `us-central1`. |
+| `model` | String | Required | The Vertex AI model ID, for example, `gemini-2.5-flash`. |
+| `auth_mode` | String | Optional | Set to `adc` to use Application Default Credentials or Workload Identity. Omit in service account key mode. |
+| `scopes` | String | Optional | The OAuth 2.0 scopes to request. Default is `https://www.googleapis.com/auth/cloud-platform`. |
```
kolchfa-aws: How do you specify multiple scopes in this parameter?

_ml-commons-plugin/remote-models/google-cloud.md
```diff
@@ -0,0 +1,150 @@
...
+    "version": 1,
+    "protocol": "google_cloud",
+    "parameters": {
+        "project_id": "<project_id>",
+        "location": "us-central1",
+        "model": "gemini-2.5-flash",
+        "auth_mode": "adc",
+        "scopes": "https://www.googleapis.com/auth/cloud-platform"
+    },
+    "credential": {},
```
kolchfa-aws: Is the empty credential required or can it be omitted?

kolchfa-aws (commented):
Thank you, @tonypiazza! Doc review complete. Content refactored to be separated by the authentication type. Could you answer the 2 outstanding questions, and we'll be ready to merge the PR.

_ml-commons-plugin/remote-models/google-cloud.md
tonypiazza: @kolchfa-aws You can't specify more than one today. The implementation wraps the raw string in a single-element list (`Collections.singletonList(connector.getScopes())` in `GoogleCredentialProvider`), so a comma- or space-separated value would be sent as one malformed scope rather than being split. In practice this isn't limiting, because the default `https://www.googleapis.com/auth/cloud-platform` already covers Vertex AI, and it's the scope every blueprint uses.

This row says "scopes," which implies plural support. Something like "The OAuth 2.0 scope to request. Only one scope is supported." would be accurate.

_ml-commons-plugin/remote-models/google-cloud.md
tonypiazza: @kolchfa-aws It can be omitted. In ADC mode the validation only rejects a `credential` that *contains* `private_key` or `client_email`; a null or absent `credential` is fine (`GoogleCloudConnectorTest.decrypt_AdcMode_LeavesNonNullDecryptedCredential` builds a valid connector with no `credential` at all). I included `"credential": {}` here to contrast with service account key mode, but it's easy to read as required.

kolchfa-aws: Thanks. Updating the docs based on this. The other parameters (backend_roles, access_mode, and add_all_backend_roles) are not required either so updating the blueprint documentation.

_ml-commons-plugin/remote-models/google-cloud.md
kolchfa-aws: Changed to singular "scope".

kolchfa-aws (approved):
Thanks, @tonypiazza! Updated. LGTM

PR #12992, opened 2026-08-28
Add doc for allowing -1 index.search.idle.after

### LINKED ISSUE #9707: Increase `index.search.idle.after` setting default from 30s to 10 minutes (or more)
**Is your feature request related to a problem? Please describe.**
Elasticsearch 7.0 introduced a "search idle" feature (https://github.com/elastic/elasticsearch/pull/27500) to avoid refreshing an index that isn't receiving any search traffic. This helps remove the unnecessary effort of refreshing during large bulk load operations. For example, for a "rebuild the index overnight and serve traffic during the day" use-case, it's apparently a big help.

Unfortunately, I've seen at least a few cases where users end up with shards going idle and then block on refresh on their next query:
1. I saw a cross-cluster replication case where the follower would receive a burst of updates during an idle period, then the next query would take a long pause.
2. If explicit routing is used, some copies of some shards are able to go idle even though the index is receiving continuous traffic, just because their specific shards haven't received traffic recently.

**Describe the solution you'd like**
In my opinion, the default 30 second shard idle timeout is far too aggressive. 

For the "big overnight re-index job" use-case or "index logs constantly and only search them when something breaks" use-case, not seeing any query traffic for 10 minutes should still be a fine threshold -- sure you're doing unnecessary refreshes for an extra 9.5 minutes, but that's not likely to be too cost prohibitive. 

**Describe alternatives you've considered**
We could change the default behavior for a search on idle shards to a background refresh, rather than blocking the first search(es). Searches could run quickly using the last (pre-idle) `IndexReader`. Unfortunately, that would be a major change for users who might be surprised by (potentially very) stale results.

We also have a workaround where users who search and update their index all the time (but sometimes have sparser search traffic) can disable the search idle feature altogether by explicitly setting `index.refresh_interval`. In my opinion, it's still a good idea to do that, but I'd like the default behavior to be less aggressive.

**Additional context**
N/A

tonklable:
### Description
Currently, OpenSearch sets index.search.idle.after to 30s by default. To disable idle time, there is only one way, to set index.refresh_interval explicitly. However, index.refresh_interval is 1s by default, and we did not want to change it. It is not a good convention to set one variable explicitly for a different behavior (disabling idle).

-1 is a standard convention to disable something, so this PR is to allow -1 for index.search.idle.after in order to disable idle time.

Related to https://github.com/opensearch-project/OpenSearch/pull/22873

### Issues Resolved
Closes https://github.com/opensearch-project/OpenSearch/issues/9707#issuecomment-5368306320

### Version
_List the OpenSearch version to which this PR applies, e.g. 2.14, 2.12--2.14, or all._

### Frontend features

### Checklist

### DIFF
--- a/_install-and-configure/configuring-opensearch/index-settings.md
+++ b/_install-and-configure/configuring-opensearch/index-settings.md
@@ -314,7 +314,7 @@ OpenSearch supports the following dynamic index-level index settings:
 
 - `index.blocks.write` (Boolean): Specifies whether the index is read-only. Setting to `true` blocks all write requests and makes the index read-only. Default is `false`.
 
-- `index.search.idle.after` (Time unit): The amount of time a shard should wait for a search or get request until it goes idle. Default is `30s`.
+- `index.search.idle.after` (Time unit): The amount of time a shard should wait for a search or get request until it goes idle. Can be set to -1 to disable idle. Default is `30s`.
 
 - `index.search.default_pipeline` (String): The name of the search pipeline that is used if no pipeline is explicitly set when searching an index. If a default pipeline is set and the pipeline doesn't exist, then the index requests fail. Use the pipeline name `_none` to specify no default search pipeline. For more information, see [Default search pipeline]({{site.url}}{{site.baseurl}}/search-plugins/search-pipelines/using-search-pipeline/#default-search-pipeline).
 


kolchfa-aws:
Thank you for this PR, @tonklable! When the code PR gets merged, please list the version that this will go into and ping one of the code reviewers to review this PR. Thanks!

PR #12994, opened 2026-08-30
Add LibreDB Studio to the Tools page

kaya-abdullah:
### Description
This adds LibreDB Studio, an open-source web-based database editor, to the Tools page (`_tools/libredb-studio.md`), following the pattern of the existing entries such as Grafana and the Terraform provider. It also adds three screenshots to `images/` showing the connection dialog, a query and its results, and the monitoring view.

The description is based on testing done locally against a real OpenSearch 3.8.0 cluster with the security plugin disabled:

- Connects to an OpenSearch cluster over HTTP using the bundled SQL plugin (`POST /_plugins/_sql`). No additional client library or driver is required.
- Lists an OpenSearch cluster's indices in a schema browser, reading each index's field mappings to show the mapped fields and their OpenSearch types.
- Runs OpenSearch SQL statements from a web-based editor, including `SELECT` queries with `LIMIT n OFFSET m` paging, which the OpenSearch SQL plugin accepts.
- Shows a basic monitoring view with cluster version, total storage size, and per-index document counts and sizes.

`nav_order` for the new file is set to 220, after the current highest sibling value (Sycamore, 210), to avoid a collision.

### Issues Resolved
N/A

### Version
3.8

### Frontend features
N/A

### Checklist


kolchfa-aws:
@kaya-abdullah 

> Connects to an OpenSearch cluster over HTTP

Does this work with the Security plugin enabled?

kaya-abdullah:
Hi @kolchfa-aws,

Yes, it connects to a Security plugin enabled cluster over HTTPS with Basic auth, but there is one caveat worth calling out.

Basic auth (`user`/`password` in the connection) works correctly against a secured cluster. I set up OpenSearch 3.8.0 with the Security plugin enabled (not the disabled plugin fixture we normally test against) and confirmed:

```
$ curl -sk -u admin:<password> https://localhost:9210/_cluster/health
HTTP 200, full cluster health JSON (status: green)

$ curl -sk -u admin:<password> -XPOST https://localhost:9210/_plugins/_sql \
  -H 'content-type: application/json' -d '{"query":"SELECT 1"}'
HTTP 200
{"schema":[{"name":"1","type":"integer"}],"datarows":[[1]],"total":1,"size":1,"status":200}
```

The caveat is TLS certificate verification. The connection is made through the platform's standard `fetch`, which always verifies the certificate chain against the system trust store and has no option to trust a custom CA or skip verification (documented here: https://github.com/libredb/libredb-studio/blob/main/docs/providers/opensearch.md#43-tls). Against a stock secured node's default self signed certificate, this means the connection fails before auth is even reached:

```
$ curl -u admin:<password> https://localhost:9210/_cluster/health
curl: (60) SSL certificate problem: unable to get local issuer certificate
```

I reproduced the same failure through the actual connection code, independent of curl:

```
ConnectionError: Failed to connect to OpenSearch: OpenSearch could not be reached:
unable to verify the first certificate: UNABLE_TO_VERIFY_LEAF_SIGNATURE
```

This happens the same way regardless of which TLS mode is selected in the connection form (`require` and `verify-system` both fail identically), since the transport does not branch on the mode.

Plain HTTP is not a fallback either. With the Security plugin enabled, a plain HTTP request to the same port is refused at the socket level (`ECONNRESET`), not silently allowed.

Once the cluster's certificate is trusted by the client (a CA signed certificate, a terminating reverse proxy, or trusting the CA explicitly for a local test), the same connection and query succeed immediately with no other change:

```
{"rows":[{"1":1}],"fields":["1"],"rowCount":1,"executionTime":7,"columnTypes":{"1":"integer"}}
```

Short version: it works with the Security plugin enabled and Basic auth, as long as the cluster presents a certificate the client already trusts. It does not work out of the box against the plugin's default self signed certificate. This matches our connection field and TLS documentation for this database type: https://github.com/libredb/libredb-studio/blob/main/docs/providers/opensearch.md#41-configuration-fields and https://github.com/libredb/libredb-studio/blob/main/docs/providers/opensearch.md#43-tls.

Let me know if you think this caveat should be called out on the Tools page itself.

kolchfa-aws:
Thanks, @kaya-abdullah, the limitations are understood. That said, the tools we document in the Tools section (Logstash, Grafana, the Terraform provider) each require OpenSearch-specific instructions that users can't get anywhere else. LibreDB Studio is a client of the SQL plugin's `POST /_plugins/_sql` endpoint, which is already documented in the [SQL and PPL](https://docs.opensearch.org/latest/search-plugins/sql/index/) section, so there's no OpenSearch-side material that we need to explain to users. 

I'm going to close this doc PR, but this looks like it can be added to the community projects page here https://opensearch.org/community-projects/. If you'd like, please follow the instructions on that page to submit the project for a review.

PR #13012, opened 2026-09-08
Update wildcard field documentation - add storage requirements and limitations

kolchfa-aws:
Add a storage requirements section to the wildcard query documentation, expand the parameter table, and add an example storage calculation.

### Checklist


kolchfa-aws:
@msfroh This PR is in response to doc readers asking to talk about storage requirements for wildcard fields. Could you please take a look at it when you get a chance: this is what I've been able to reconstruct from the code. Thank you!

_mappings/supported-field-types/wildcard.md
```diff
@@ -45,25 +45,145 @@ PUT logs
...
+
+Two properties of this measurement are worth noting. Because the indexes are merged into a single segment, the sizes are a floor: a production index holds multiple segments and deleted documents, both of which add overhead to either field type. And the ratio decreases slightly as the document count grows, so a small sample yields a marginally conservative estimate.
+
+To reduce the cost, use `ignore_above` to skip indexing values longer than a chosen length, or set `doc_values` to `false` if you do not need aggregations, sorting, or scripting on the field. When `doc_values` is disabled, OpenSearch reads each candidate document's value from `_source` rather than from doc values when it filters the substring matches, so the field must remain in `_source`. If the field is excluded from `_source` and `doc_values` is disabled, these queries return no matches instead of an error. Either setting makes the field ineligible for [derived source](#derived-source).
```
msfroh: Disabling doc values on `wildcard` fields might make the queries considerably slower as fetching from `_source` tends to be much slower than reading from doc values. (We need to double-check each possible match against doc values or `_source`, since the index lookup can produce false positives.)

_mappings/supported-field-types/wildcard.md
```diff
@@ -45,25 +45,145 @@ PUT logs
...
+
+## Limitations
+
+The following queries are not supported on `wildcard` fields:
+
+- [`fuzzy`]({{site.url}}{{site.baseurl}}/query-dsl/term/fuzzy/) queries: Run them on a `keyword` or `text` field instead.
+- [`range`]({{site.url}}{{site.baseurl}}/query-dsl/term/range/) queries: Range matching does not apply to this field type.
```
msfroh, line 180: I feel like both of these should probably be supported eventually (if anyone expresses interest in them).

In particular, I can see how to do an efficient implementation for range queries (since they're just a generalization of prefix queries). For fuzzy queries, we could just ignore the indexed trigrams and apply the fuzzy automaton to doc values -- it stinks if that's your only clause, but if you just need to filter an already reduced match set based on fuzzy matching, it could be okay.

msfroh (approved):

_mappings/supported-field-types/wildcard.md
kolchfa-aws: Thank you! Updated the wording.

peterzhuamazon (approved):

PR #13026, opened 2026-09-09
Redirect the home page to latest and update canonical links in 1.3

kolchfa-aws:
### Description

The June 2024 pass (#7517, #7535 to #7550) already redirected every Clients, Data Prepper, Benchmark, and Migration Assistant page on this branch, so nothing is left to do there. What remains is the branch home page.

`/1.3/` serves a home page frozen at that release. GA shows no versioned root page has a single referrer from `docs.opensearch.org`, against 37% on-site referrals for `/latest/`, so nobody navigates within a version from its home page. The arrivals are external: bookmarks, third-party links, and OpenSearch Dashboards, whose `noDocumentation` block holds 62 version-pinned links that resolve to exactly this root. All of them are better served by the current home page than a stale one. Content pages are out of scope: a reader who lands on `/1.3/dashboards/dql/` stays in 1.3.

`_plugins/link-checker.rb` gains `^/$` in `@ignored_paths`. CI builds every branch with `baseurl: /latest`, so without it the checker rewrites the home page's redirect target back to `/`, reads the stub it just generated, pulls the same absolute URL out of it, and recurses until the build dies with `stack level too deep`. That was the cause of the CI failures on the first attempt at this pass.

Verified locally with `JEKYLL_FATAL_LINK_CHECKER=internal` on 2.13 and 3.7, one branch of each `@ignored_paths` shape. Both report no broken links, and `_site/index.html` is the redirect stub.

`@ignored_paths` ends up covering the home page and all five single-edition section paths on this branch, so nothing that carries a `redirect_to` is followed:

```ruby
@ignored_paths = %r{(^/$|^/javadocs|^mailto:|^/clients|^/data-prepper|^/benchmark|^/migration-assistant|^/classic/migration-assistant)}.freeze
```

The pattern matches the link target rather than the page the link sits on, so these entries also stop inbound links to those sections from being verified. That costs nothing here: the targets are redirect stubs the checker would skip anyway, and no new content will land on this branch. It would cost real coverage on `main`, where those are live pages, so `main` keeps a shorter list on purpose (#13056) and this line is not meant to be synced to it.

[...]

PR #13056, opened 2026-09-10
Keep the link checker's ignored paths identical on every branch

kolchfa-aws:
### Description

The redirect passes on 1.3 through 3.7 (#13026 to #13054) point each branch's home page and its Clients, Data Prepper, Benchmark, Migration Assistant, and Migration Assistant (Classic) pages at `/latest/`. Those redirects need matching entries in `@ignored_paths`, because CI builds every branch with `baseurl: /latest`: the checker rewrites the redirect target back to the page's own path, reads the stub `jekyll-redirect-from` just generated, pulls the same absolute URL back out of it, and recurses until the build dies with `stack level too deep`. `check_internal` has no cycle detection.

`main` had none of those entries. This adds all of them, in the same order and with the same comment, so the line is byte-identical on 1.3 through `main`. Two reasons to carry it here rather than only on the version branches:

- A branch cut from `main` already has the fix. Today the release process has to patch it in later, once the branch stops being the source of `/latest/` and the redirects go on.
- One form of the line means a future change to it is one change, not thirty.

Every entry is a no-op on `main`, which is where `/latest/` is built and where all of these pages are real content rather than redirect stubs.

**One tradeoff worth flagging:** internal links pointing *into* those five sections are no longer verified on `main`. On a version branch nothing is lost, since the targets are stubs the checker skips anyway, but on `main` they are live pages, and `main` is where a broken link is most likely to be introduced. A typo like `/clients/javascpirt/` would now pass. The alternative is to keep `main` on a shorter list and accept that every new release branch needs the longer one patched in by hand.

Verified locally with `JEKYLL_FATAL_LINK_CHECKER=internal` on 2.13 and 3.7, one branch of each prior `@ignored_paths` shape. Both report no broken links.

### Issues Resolved

N/A

### Version

main

### Frontend features

N/A

### Checklist

### DIFF
--- a/_plugins/link-checker.rb
+++ b/_plugins/link-checker.rb
@@ -91,8 +91,20 @@ module Jekyll::LinkChecker
   ]
 
   ##
-  # Pattern of local paths to ignore
-  @ignored_paths = %r{(^/javadocs|^mailto:)}.freeze
+  # Pattern of local paths to ignore.
+  #
+  # `^/$` is the site home page, and the five section paths are the sections
+  # published as a single edition. On a branch that is no longer the source of
+  # /latest/, every one of those pages carries a `redirect_to` aimed at /latest/.
+  # CI sets `baseurl: /latest`, so the checker rewrites that target back to the
+  # page's own path, reads the stub jekyll-redirect-from just generated, pulls the
+  # same absolute URL out of it, and recurses until the build dies with `stack
+  # level too deep`. `check_internal` has no cycle detection.
+  #
+  # Every entry is a no-op on whichever branch builds /latest/, where these pages
+  # are real content. The list is kept identical on all branches so that one cut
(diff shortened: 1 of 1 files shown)


kolchfa-aws:
Closing this. Carrying the full `@ignored_paths` list on `main` costs more than it saves.

`@ignored_paths` is matched against the link target, not the page the link is on, so `^/clients` and friends would skip every inbound link to those sections from anywhere in the docs, plus every link to the home page. On a version branch that costs nothing, because the targets are redirect stubs the checker skips anyway and no new content lands there. On `main` they are live pages, and `main` is where new links are written — so this would turn off exactly the checking that catches a typo like `/clients/javascpirt/`.

Adding the entries at release time instead, in the same PR as the redirects, where each one is genuinely a no-op. Written up as a step in the release skill.

PR #13059, opened 2026-09-10
Refactor the Security API section, add a Managing passwords page, and add a security option to the Copy as cURL button

kolchfa-aws:
### Description

This PR reorganizes the Security plugin documentation around a new `_security/api/` section, adds a page that explains the different passwords in a cluster, and teaches the **Copy as cURL** button to emit a TLS-aware command.

**1. Security option for the Copy as cURL button**

`{% include copy-curl.html %}` now accepts a `security` flag. On a page documenting an API that requires the Security plugin, `{% include copy-curl.html security=true %}` makes the button produce a command that works against a secure cluster:

```
curl -k -u 'admin:<custom-admin-password>' -XGET "https://localhost:9200/_plugins/_security/api/roles"
```

Instead of the default:

```
curl -XGET "http://localhost:9200/_plugins/_security/api/roles"
```

Pages that don't pass the flag are unaffected.

The same change fixes a pre-existing bug in the cURL conversion: the body was detected with the first `{` anywhere in the snippet, so a request whose path contained `{` (or a JSON array body starting with `[`) produced a malformed command. Body detection now starts after the request line and accepts either `{` or `[`.

**2. Refactored Security API section**

The Security API reference was split across two places: a small `_api-reference/security/` tree and a 2,000-line monolithic page at `_security/access-control/api.md`. Both are now consolidated into a single `_security/api/` section with one page per operation — 77 pages across 15 API groups plus 3 standalone pages:

Authentication, Account, Internal user, Role, Role mapping, Action group, API key, Tenant, Multi-tenancy configuration, Allow list, Configuration, Security configuration version, Audit log, Distinguished name, and Certificate APIs, plus the Flush Cache, Dashboards Info, and Security Plugin Health APIs.

Details:

- Every page follows the API reference template: endpoints, path/query/request body parameters, example request, example response, and response body fields. Endpoints and parameter tables come from `spec_insert` blocks where the spec is accurate; where the spec was wrong or incomplete, the tags were removed and the table written by hand.
- All requests and responses were tested against a local OpenSearch 3.8.0 cluster with the Security plugin enabled, and the example responses are actual output.
- Testing turned up several undocumented or incorrectly documented behaviors, which are now corrected. For example, the Authorization Token API returns an empty `message` and does not issue a token; the Generate User Token API works only for service accounts, which must be created without a password; the SSL Info API no longer returns the six `ssl_openssl_*` fields; and the Create or Update Multi-Tenancy Configuration API accepts a `preferred_tenants` field that wasn't documented.
- `_security/access-control/api.md` is retained as **API permissions**, covering only access control for the API, `plugins.security.restapi.*` settings, and reserved/hidden resources. It links to the new reference for endpoints.
- `_security/configuration/versioning.md` is folded into the Security configuration version APIs group.
- ~15 pages in other collections that deep-linked into the old monolithic page (`.../access-control/api#create-user` and similar) now link to the corresponding new page.
- The Configuration section's `nav_order` values were renumbered to a consistent scale to accommodate the new page and the moved versioning content.

**3. New Managing passwords page**

A cluster running the Security plugin has several distinct passwords, each set by a different person, stored in a different place, and changed through a different procedure. Password guidance was scattered across the Docker install page, the demo configuration page, and `yaml.md`, and the admin password requirements were duplicated in six install pages.

`_security/configuration/passwords.md` is a new page that lists all of them in one table — admin, internal user, your own, the Dashboards service account, and the keystore/truststore passwords — with who sets each one, where it's stored, and how to change it. The admin password requirements now live in one place, in the demo configuration page, and the install pages link to it.

### Issues Resolved

N/A

### Version

all

### Frontend features

N/A

### Checklist


_security/api/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+Four APIs are exempt because they expose only the requesting user's own information: the [account APIs]({{site.url}}{{site.baseurl}}/security/api/account/), the [Permissions Info API]({{site.url}}{{site.baseurl}}/security/api/authentication/permissions-info/), the [Dashboards Info API]({{site.url}}{{site.baseurl}}/security/api/dashboards-info/), and the [Security Plugin Health API]({{site.url}}{{site.baseurl}}/security/api/health/). Any authenticated user can call them.
+
+Three groups of APIs require more than a `plugins.security.restapi.roles_enabled` role. The [allow list APIs]({{site.url}}{{site.baseurl}}/security/api/allowlist/), [distinguished name APIs]({{site.url}}{{site.baseurl}}/security/api/distinguished-names/), and [certificate APIs]({{site.url}}{{site.baseurl}}/security/api/certificates/) are restricted to a super admin, and each category page describes the additional cluster permission that lets a role reach them. For more information, see [Access control for the API]({{site.url}}{{site.baseurl}}/security/access-control/api/).
+
+Most Security API calls require HTTP basic authentication with admin credentials, as shown in the following example:
```
cwperks: nit: they don't require basic auth per se. The user calling the APIs needs one of the roles from `plugins.security.restapi.roles_enabled: [...]`

cwperks (approved):

_security/api/index.md
kolchfa-aws:
```suggestion
Most Security APIs require the calling user to be mapped to a role listed in `plugins.security.restapi.roles_enabled`. The following example uses HTTP basic authentication as the `admin` user:
```

kolchfa-aws: Thanks - reworded

PR #12938, opened 2026-08-18
Add dynamic mapping for knn_vector fields

naykudev:
## Description

Documents the two dynamic `knn_vector` mapping paths introduced in 3.9:

- **Dynamic templates** — using `knn_vector` as a `match_mapping_type`, with `dimension` either specified or inferred from the first document.
- **Auto-inference** — zero-configuration mapping of an unmapped flat numeric array as a `knn_vector` when its length is a multiple of 8 and within the `[128, 16000]` range.

The section also covers dimension locking after the first document, the shape-based nature of auto-inference (including the misclaim caveat for non-vector numeric arrays), and that an explicit mapping always takes precedence.

Added as a new "Dynamic mapping" section in `_mappings/supported-field-types/knn-vector.md`.

## Related

- Feature (core): opensearch-project/OpenSearch#22607
- Feature (k-NN): opensearch-project/k-NN#3490

## Checklist
- [x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #13082, opened 2026-09-16
Add dynamic mapping for knn_vector fields

naveentatikonda:
## Description

Supersedes #12938 (opened by @naykudev). Rebased onto latest `main` and expanded to reflect the feature as merged in [opensearch-project/k-NN#3490](https://github.com/opensearch-project/k-NN/pull/3490).

Documents the two dynamic `knn_vector` mapping paths introduced in 3.9:

- **Dynamic templates** — using `knn_vector` as a `match_mapping_type`, with `dimension` inferred from the first document (or preserved when `dimension` or `model_id` is specified). The `type: knn_vector` field is optional inside the template `mapping` block and is injected automatically.
- **Auto-inference** — an unmapped flat numeric array is mapped as `knn_vector` when its length is a multiple of 8 and within `[128, MAX_DIMENSION]` (16,000 for the default Faiss engine).

Additional context added on top of #12938:

- The feature is gated by the new cluster setting `knn.dynamic_mapping.enabled` (default `false`). Added the setting entry to `_vector-search/settings.md` and referenced it from the Dynamic mapping section.
- Clarified that dynamic mapping creates the `knn_vector` field regardless of `index.knn`, so exact k-NN search always works — but **approximate (ANN) search requires `index.knn: true`**, which is a static index setting that must be set at index-creation time and cannot be enabled later. Highlighted as a warning callout so users don't create an index expecting ANN and get locked into an exact-only field.
- Clarified that dynamic mapping applies to new indices only and to fields that are not already mapped.
- Clarified that the array-length heuristic used by auto-inference does not apply to explicit dynamic templates.
- Reworked the auto-inference example so it is valid JSON (removed inline `//` comment).

## Related

- Feature (core): opensearch-project/OpenSearch#22607
- Feature (k-NN): opensearch-project/k-NN#3490 (merged)
- Supersedes: #12938

## Checklist


kotwanikunal (approved):

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
 `method` | Object | The algorithm used for organizing vector data at indexing time and searching it at search time. Used when the ANN algorithm does not require training. Optional. For more information, see [Methods and engines]({{site.url}}{{site.baseurl}}/mappings/supported-field-types/knn-methods-engines/). 
 `model_id` | String | The model ID of a trained model. Used when the ANN algorithm requires training. See [Model IDs](#model-ids). Optional.
 
+## Dynamic mapping
+**Introduced 3.9**
+{: .label .label-purple }
+
+You can have OpenSearch map a field as a `knn_vector` automatically, without declaring it in your mappings up front. There are two paths: a dynamic template that uses `knn_vector` as a `match_mapping_type`, and auto-inference from the first indexed value.
+
+Dynamic mapping applies only to new indexes and only to fields that are not already mapped. An explicit mapping always takes precedence: if a field is already mapped, neither path runs for it.
```
kolchfa-aws: What does "applies to new indexes" mean? Does it mean that for existing indexes, if you index a new field that is a k-NN vector, it's not going to be dynamically mapped? Also, the second sentence applies to all field types, correct? So it doesn't convey new information.

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
...
+      "knn": true
+    }
+  },
+  "mappings": {
+    "dynamic_templates": [
+      {
+        "vectors": {
+          "match_mapping_type": "knn_vector",
+          "mapping": {
+            "type": "knn_vector"
```
kolchfa-aws, line 222: Since type: knn_vector is optional, should we have "mapping": {} here?

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
...
+          }
+        }
+      }
+    ]
+  }
+}
+```
+{% include copy-curl.html %}
+
```
kolchfa-aws, line 231: Can we add an example that indexes a document after this is in place and a request for the users to confirm that the mapping was performed correctly with a response?

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
...
+You can specify any `knn_vector` parameters (such as `dimension`, `space_type`, `method`, or `model_id`) in the `mapping` block. If you specify `dimension`, or a `model_id` that supplies it, OpenSearch uses that value directly and skips inference from the first document. If you omit both, the dimension is inferred from the first indexed document, after which it is fixed for the field.
```
kolchfa-aws: So, to confirm, a template-mapped vector isn't required to be >=128 or a multiple of 8 so a vector [0.1, 0.2, 0.3, 0.4] maps as knn_vector with dimension: 4?

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
...
+
+- If `index.knn: true`, OpenSearch builds the data structures required for the field, and both exact and approximate k-NN search are supported.
+- If `index.knn` is unset or `false`, the field is still mapped as `knn_vector`, but only exact k-NN search is supported.
+
+If you plan to run ANN search on auto-inferred vector fields, create the index with `index.knn: true` up front. You cannot enable ANN search on an existing index---you must reindex into a new index created with `index.knn: true`.
+{: .warning}
+
+### Dynamic templates
+
+You can reference `knn_vector` as the `match_mapping_type` in a [dynamic template]({{site.url}}{{site.baseurl}}/field-types/#dynamic-mapping). When a matching field is first encountered, OpenSearch maps it as a `knn_vector` using the mapping block you provide. If you omit `dimension`, it is inferred from the length of the first indexed vector:
```
kolchfa-aws: What defaults does a dynamically mapped vector get? The default engine, space type, and method?

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
...
 
+## Dynamic mapping
+**Introduced 3.9**
+{: .label .label-purple }
+
+You can have OpenSearch map a field as a `knn_vector` automatically, without declaring it in your mappings up front. There are two paths: a dynamic template that uses `knn_vector` as a `match_mapping_type`, and auto-inference from the first indexed value.
+
+Dynamic mapping applies only to new indexes and only to fields that are not already mapped. An explicit mapping always takes precedence: if a field is already mapped, neither path runs for it.
+
+Dynamic mapping is disabled by default. To enable it, set the [`knn.dynamic_mapping.enabled`]({{site.url}}{{site.baseurl}}/vector-search/settings/#cluster-settings) cluster setting:
```
kolchfa-aws, line 184: Does dynamic knn_vector mapping require setting `dynamic: true` on the index or object?

_mappings/supported-field-types/knn-vector.md
```diff
@@ -173,6 +173,112 @@ Parameter | Data type | Description
...
+- A numeric array that is not a vector (for example, a large list of IDs or measurements) is mapped as a `knn_vector` if its length happens to meet these conditions. Because the dimension is fixed after the first document, later documents whose array has a different length are rejected. To prevent a field from being auto-inferred as a `knn_vector`, declare an explicit mapping for it or use a dynamic template that maps the field to a different type.
+
+### Limitations
+
+Dynamic mapping does not create [`nested`]({{site.url}}{{site.baseurl}}/field-types/supported-field-types/nested/) parents. If you want to index an array of objects where each object contains its own vector (for example, per-chunk embeddings for a document), you must still declare the parent field explicitly as `type: nested`. Auto-inference and dynamic templates apply only to individual `knn_vector` fields, not to their parent object type---an array of objects such as the following is mapped as a plain `object`, not `nested`, so per-inner-document vector search does not work:
+
+```json
+{
+  "chunks": [
```
kolchfa-aws, line 318: What happens at index time when one document supplies several values for that field - is the document rejected or still indexed?

kolchfa-aws (commented):
Thank you, @naveentatikonda! I have a couple of clarifying questions before I start rewriting - could you please take a look?

_mappings/supported-field-types/knn-vector.md
naveentatikonda: actually it applies to all `knn_vector` fields which are not yet mapped even though we are adding those fields on indices that were created in older versions. I will fix it in documentation.

> the second sentence applies to all field types, correct?

All `knn_vector` fields only

_mappings/supported-field-types/knn-vector.md
naveentatikonda: no, mapping block is mandatory

_mappings/supported-field-types/knn-vector.md
naveentatikonda: The inferencer sets only type: knn_vector and dimension: N. It will pick other defaults based upon the k-NN mapper logic. For example :
```
- engine: faiss (KNNEngine.DEFAULT)                                                                                                                                                                                                    
  - method.name: hnsw                                                                                                                                                                                                                    
  - space_type: l2                                                                                                                                                                                                                       
  - data_type: float
  ```

File: _mappings/supported-field-types/knn-vector.md (hunk 6 above)
Comment (naveentatikonda) [reply _mappings/supported-field-types/knn-vector.md:184]: Yes, k-NN dynamically infer field mappings just like how we map other dynamic fields like `text` by following the same dynamic settings - https://docs.opensearch.org/latest/mappings/mapping-parameters/dynamic/

File: _mappings/supported-field-types/knn-vector.md (hunk 4 above)
Comment (naveentatikonda) [reply _mappings/supported-field-types/knn-vector.md]: yes, because they are explicitly setting that as a knn_vector so we are just inferring the dimension from 1st document, that too if they didn't specify dimension in mapping. 

The 128 dimension threshold isn't required for dynamic templates

File: _mappings/supported-field-types/knn-vector.md (hunk 7 above)
Comment (naveentatikonda) [reply _mappings/supported-field-types/knn-vector.md:318]: If they don't define parent field as nested and try to ingest nested vectors then the ingestion will fail

File: _mappings/supported-field-types/knn-vector.md (hunk 3 above)
Comment (naveentatikonda) [reply _mappings/supported-field-types/knn-vector.md:231]: Added

File: _mappings/supported-field-types/knn-vector.md (hunk 6 above)
Comment (kolchfa-aws) [reply _mappings/supported-field-types/knn-vector.md:184]: Follow-up question: what happens if `dynamic` is set to `strict`?

File: _mappings/supported-field-types/knn-vector.md [hunk 8]
```diff
@@ -173,6 +173,215 @@ Parameter | Data type | Description
...
+```
+{% include copy.html %}
+
+Because the template establishes the field type, the array-length heuristic used by auto-inference (described in the following section) does not apply. Any flat numeric array that matches the template is mapped as a `knn_vector`, regardless of its length.
+
+### Auto-inference
+
+When no dynamic template matches, OpenSearch can still infer a `knn_vector` mapping from the field value. An unmapped field is mapped as a `knn_vector` when its value is a flat array of numbers whose length is a multiple of 8 and falls within the range from 128 to the maximum dimension supported by the default k-NN engine (16,000 for Faiss). This bound is applied at inference time regardless of which engine the field ultimately uses. The dimension is set to the array length. An array whose length falls outside this range or is not a multiple of 8 is mapped as a numeric array.
+
+For example, `knn-auto-infer` has no mapping for `embedding` and no dynamic template. Indexing the following document maps `embedding` as a `knn_vector` of dimension 768. The array is truncated in this example:
```
kolchfa-aws, line 301: For this example, knn-auto-infer is never created so if a reader is trying out this request, the index that's auto-created will be the one created with POST. Could you add a PUT /knn-auto-infer with index.knn: true before the POST? That would also let this example end with a GET /knn-auto-infer/_mapping showing dimension: 768, matching the structure of the dynamic template and nested examples.

kolchfa-aws (commented):
Thanks for the clarifications, @naveentatikonda! Doc review complete and changes applied. Could you take a look at the changes and also answer a couple of follow-up questions?

naveentatikonda:
> Follow-up question: what happens if dynamic is set to strict?

@kolchfa-aws the document is rejected during ingestion in core's DocumentParser

_mappings/supported-field-types/knn-vector.md
naveentatikonda: irrespective of the index setting the mapping will be inferred but ANN search won't work

naveentatikonda:
> Thanks for the clarifications, @naveentatikonda! Doc review complete and changes applied. Could you take a look at the changes and also answer a couple of follow-up questions?

@kolchfa-aws all the changes looks good. Thanks!

kolchfa-aws (approved):
Thank you, @naveentatikonda! Updated based on your answers. LGTM

PR #13123, opened 2026-09-17
Clarify that config.enabled in audit.yml does not enable audit logging on its own

### LINKED ISSUE #12696: [DOC] Audit log is documented as disabled by default, but is not
**What do you want to do?**
 
- [ ] Add new documentation
- [x] Report a technical problem with the documentation
- [ ] Other 

https://docs.opensearch.org/latest/security/audit-logs/index/ states that audit logs are disabled by default. However, the `audit.yml` file provided by the opensearch-security plugin has audit logs enabled. 

https://github.com/opensearch-project/security/blob/76c3e2c48d7e88fd9c82dc6cff739d6e7098a326/config/audit.yml#L7

```
root@default:/home/vagrant# cat /etc/opensearch/opensearch-security/audit.yml 
_meta:
  type: "audit"
  config_version: 2

config:
  # enable/disable audit logging
  enabled: true
```

**Version:** Latest

gohri81-tech:
### Description

Closes #12696, where the shipped `audit.yml` (which has `config.enabled: true`) was read as a contradiction of the statement on this page that audit logging is disabled by default.

The page is right, and the reason is worth spelling out, so this adds one note after the enabling steps. From the Security plugin source:

- `SinkProvider` builds the default sink from `plugins.security.audit.type`, and `createSink` returns `null` when that type is `null`, logging `Default endpoint could not be created, auditlog will not work properly`.
- `AuditMessageRouter.isEnabled()` is `defaultSink != null`, and it warns `No default storage available, audit log may not work properly` when there is none.
- `AuditLogImpl.setConfig` computes `enabled = auditConfig.isEnabled() && messageRouterEnabled`, so `config.enabled: true` in `audit.yml` is only one of the two conditions.

So a default install has `config.enabled: true` and no storage type, which is why nothing is recorded and why the startup log complains about storage rather than about being disabled. That second part is what sent the reporter looking (see the linked wazuh/wazuh-indexer#1587).

### Issues Resolved

Closes #12696

### Version

all

### Checklist

- [x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and subject to the Developer Certificate of Origin.

One note on the diff: the web editor rewrote the final `{% raw %}{% include copy.html %}{% endraw %}` line because the file has no trailing newline, so that shows up as one addition and one deletion with no visible change. Happy to redo the change without it if you would rather keep the diff to the one paragraph.


DarshitChanpura:
Thanks for digging into the source for this one — I checked the three claims in the description against `main` and they all hold:

- `AuditLogImpl.java:89` — `enabled = auditConfig.isEnabled() && messageRouterEnabled`
- `AuditMessageRouter.java:72-73` — `isEnabled()` returns `defaultSink != null`
- `SinkProvider.java:85` — logs `Default endpoint could not be created, auditlog will not work properly.` when no default sink can be built

And the shipped `config/audit.yml` does set `config.enabled: true`, so the confusion in #12696 is reproducible exactly as described. The note reads accurately to me.

One merge-order heads-up for whoever picks this up: #12922 (standalone audit logging docs) edits the paragraph immediately below this insertion point and adds its own `#### Enable or disable audit logging` section built around the `plugins.security.audit.enabled` node setting. That setting is a different switch from `config.enabled` in `audit.yml` — it is never read at startup (`AuditConfig.from(Settings)` hardcodes `true`) and only has a runtime consumer in standalone mode — so the two PRs are not in conflict on the code, but they will conflict textually in this hunk, and the FGAC-facing explanation should end up being the one in this PR. I have left a note on #12922 suggesting it rebase onto this.

No change requested here from my side.

Re: the trailing-newline line in the diff — adding the missing newline at EOF is a fine thing to leave in; no need to redo the PR to strip it.

_security/audit-logs/index.md
```diff
@@ -39,6 +39,9 @@ Audit logging is disabled by default. To enable audit logging:
 
 2. Restart each node.
 
+The `audit.yml` file that ships with the Security plugin sets `config.enabled` to `true`, so that file on its own can make it look as though audit logging is already running. It is not. Audit logging needs both switches: a storage type in `opensearch.yml` and `config.enabled: true` in `audit.yml`. Without `plugins.security.audit.type`, the plugin cannot create a storage endpoint for the audit log, warns at startup that no default storage is available, and records no events.
```
kolchfa-aws:
```suggestion
Audit logging requires two settings: a storage type (`plugins.security.audit.type`) in `opensearch.yml` and `config.enabled: true` in `audit.yml`. The `audit.yml` file provided by the Security plugin sets `config.enabled` to `true` by default, so audit logging can appear to be enabled in a new cluster. Until you specify a storage type, the Security plugin cannot create a storage endpoint for the audit log, so it does not record any events and logs a warning at startup that no default storage is available.
```

kolchfa-aws (approved):
Thank you, @gohri81-tech! Doc review complete and edits applied.

PR #13126, opened 2026-09-18
Use the current transport hostname verification setting in the self-signed certificate example

gohri81-tech:
### Description

The sample `opensearch.yml` under "Add certificate files to opensearch.yml" on the self-signed certificates page still sets

```yml
plugins.security.ssl.transport.enforce_hostname_verification: false
```

That key is deprecated. In `OpenSearchSecuritySSLPlugin` the security plugin declares it with `Property.Deprecated`, and it only registers it at all when the current key is absent:

```java
private static final Setting<Boolean> SECURITY_SSL_TRANSPORT_ENFORCE_HOSTNAME_VERIFICATION = Setting.boolSetting(
    SECURITY_SSL_TRANSPORT_ENFORCE_HOSTNAME_VERIFICATION_KEY,
    true,
    Property.NodeScope,
    Property.Filtered,
    Property.Deprecated
);
...
if (!settings.stream().anyMatch(s -> s.getKey().equalsIgnoreCase(TRANSPORT_SSL_ENFORCE_HOSTNAME_VERIFICATION_KEY))) {
    settings.add(SECURITY_SSL_TRANSPORT_ENFORCE_HOSTNAME_VERIFICATION);
}
```

The current key is `transport.ssl.enforce_hostname_verification`, defined in core as `NetworkModule.TRANSPORT_SSL_ENFORCE_HOSTNAME_VERIFICATION_KEY`, registered in `ClusterSettings`, and read by `SecureNetty4Transport`. So following this page as written gets you the deprecation warning at startup:

> [plugins.security.ssl.transport.enforce_hostname_verification] setting was deprecated in OpenSearch and will be removed in a future release!

which is what was reported in opensearch-project/security#4780.

This changes the one line in the example. The rest of the documentation is already consistent with the new key, which is what makes this stand out rather than being a wider migration:

- This same page, in the SAN section, already tells you about `transport.ssl.enforce_hostname_verification`.
- `_security/configuration/tls.md` documents `transport.ssl.enforce_hostname_verification` and `transport.ssl.resolve_hostname`, and lists the two `plugins.security.ssl.transport.*` forms below them marked `(Deprecated)` with a pointer to the new names.
- The sample configs in `_security/configuration/yaml.md`, `_install-and-configure/install-opensearch/docker.md` and `_install-and-configure/configuring-opensearch/security-settings.md` all already use the new key.

After this, the deprecated spelling appears in the documentation only where it is explicitly labelled as deprecated, or where a page tells you to change it to the new one.

### Issues Resolved

Addresses the documentation half of opensearch-project/security#4780. That issue asked for the replacement to be documented or the warning to be removed; `tls.md` has since covered the reference table, so this is the last place the docs still hand a reader the deprecated key.

### Version

all

### Checklist

I cannot start a cluster from where I am working, so this is from reading `OpenSearchSecuritySSLPlugin.java` and `NetworkModule.java` rather than from watching the log line appear and disappear. Worth a second pair of eyes on one point: the deprecated key defaults to `true` and so does the new one, and neither is the demo configuration's business, so setting the new key to `false` here keeps the example behaving exactly as before.


kolchfa-aws:
@cwperks Could you please review this PR?

gohri81-tech:
Gentle nudge on this one, whenever someone has a moment. It is a one line change to the sample `opensearch.yml`, swapping the deprecated `plugins.security.ssl.transport.enforce_hostname_verification` for the current `transport.ssl.enforce_hostname_verification`, which the rest of the documentation already uses. Both keys default to `true`, so the example behaves exactly as before.

Happy to answer anything or adjust the wording - and if the assigned tech reviewer is tied up, any maintainer comfortable with the security plugin settings could confirm it just as well. Thank you.

kolchfa-aws:
@DarshitChanpura Would you be able to review this PR?

DarshitChanpura (approved):

kolchfa-aws (approved):

PR #13127, opened 2026-09-18
Document setting precedence in its own section and explain environment variables

gohri81-tech:
### Description

Addresses #9142.

That issue makes four points about how settings can be supplied. This PR covers the two that can be answered from the OpenSearch source, and leaves the other two alone rather than guessing.

**The precedence list was in the wrong place and incomplete.** It sat inside "Updating cluster settings using the API", which is not where a reader looks for it, and it listed only transient settings, persistent settings, `opensearch.yml` and defaults - with no mention of the `-E` flags that the same page documents a few sections further up. It is now its own `## Setting precedence` section, placed after the startup section and before the API section, with the flags included. The API section keeps a one-line pointer to it.

The order comes from `InternalSettingsPreparer.prepareEnvironment`, which loads `opensearch.yml` into the builder and then applies the command-line properties on top of it:

```java
output.loadFromPath(path);                        // opensearch.yml
...
initializeSettings(output, input, properties);    // output.putProperties(esSettings, ...)
```

So `-E` wins over the file, which is why it sits above `opensearch.yml` in the list.

**Environment variables needed the explanation the issue asked for.** The reporter's point was that you cannot do this:

```text
$ export discovery.type=single-node
bash: export: `discovery.type=single-node': not a valid identifier
```

"Directly in the shell environment" now says so, and gives the two things that do work: an `-E` flag, or a `${VAR}` placeholder in `opensearch.yml`. The placeholder route is real rather than folklore - `Settings.Builder.replacePropertyPlaceholders()` delegates to `replacePropertyPlaceholders(System::getenv)`, so placeholders are resolved from the environment as the file is read.

"Docker environment variables" now explains why `docker run -e discovery.type=single-node` works when `export` does not: the image entrypoint loops over `env` and rewrites anything setting-shaped into a flag.

```bash
if [[ "$envvar_key" =~ ^[a-z0-9_]+\.[a-z0-9_]+ || "$envvar_key" == "processors" ]]; then
  if [[ ! -z $envvar_value ]]; then
    opensearch_opt="-E${envvar_key}=${envvar_value}"
```

That is where the "at least two dot-separated lowercase words, plus `processors`" wording and the empty-value caveat come from.

### Not in this PR

Two of the four points in the issue are deliberately left open:

- **How to see where a setting came from.** The reporter was unsure whether this exists, and so am I. `GET _cluster/settings?include_defaults=true` separates persistent, transient and default values, but I could not find anything that attributes a value to `opensearch.yml` as opposed to an `-E` flag, and I would rather not document an API that does not exist. If a maintainer knows of one, I will add it.
- **How to unset a setting that is set at a lower level.** The page already covers resetting a cluster setting with `null`, but not how to remove a value coming from `opensearch.yml`. An `-E` flag can replace it, and an empty environment variable in Docker is skipped rather than applied, so there is no general answer I can state with confidence - least of all for list values, which is the case the reporter specifically asked about.

### Issues Resolved

Addresses #9142. I have not written "Closes" because of the two points above, so please close it or leave it open as you see fit.

### Version

all

### Checklist

I cannot build the site or start a cluster from where I am working, so the wording comes from reading `InternalSettingsPreparer.java` and `Settings.java` in `OpenSearch` and the Docker entrypoint in `opensearch-build`, rather than from observing the behaviour. Two things worth a reviewer's eye: the `${VAR}` example is the one place a placeholder-looking construct sits in a code block, so please confirm it renders as intended; and the shell error is in a `text` block with no copy button on purpose, since it is an illustration of a failure rather than something to run.


kolchfa-aws:
@peterzhuamazon Would you be able to review this PR please?

gohri81-tech:
A gentle reminder on this one when a reviewer has time. To make it easier to check, the two claims that need verifying are both traceable to the source: the precedence order comes from `InternalSettingsPreparer.prepareEnvironment`, which loads `opensearch.yml` and then applies the command line properties on top of it, and the `${VAR}` placeholder route comes from `Settings.Builder.replacePropertyPlaceholders()`, which resolves from `System::getenv`.

I have deliberately left the other two points of #9142 out rather than guess, so the issue can stay open if you prefer. Happy to reword anything that is not how you would put it. Thank you.

kolchfa-aws:
@reta Do you have time to take a look and review this PR please?

_install-and-configure/configuring-opensearch/index.md
```diff
@@ -47,6 +47,21 @@ export OPENSEARCH_PATH_CONF="/etc/opensearch"
...
 {% include copy.html %}
 
+Most OpenSearch settings cannot be exported this way. Their names contain dots, and most shells do not accept a dot in a variable name:
+
+```text
+$ export discovery.type=single-node
+bash: export: `discovery.type=single-node': not a valid identifier
+```
+
+The variables you can export directly are the ones the startup scripts read, such as `OPENSEARCH_JAVA_OPTS` and `OPENSEARCH_PATH_CONF`. To supply any other setting from the environment, either pass it as an `-E` flag or reference the variable from `opensearch.yml` with a `${VAR}` placeholder:
```
reta, line 57:
```suggestion
To supply such settings from the environment, either pass it as along with `-E` flag or introduce custom environment variables that could be referenced from `opensearch.yml` using `${ENV_VAR}` placeholder:
```

_install-and-configure/configuring-opensearch/index.md
```diff
@@ -77,6 +92,27 @@ docker run -e "OPENSEARCH_JAVA_OPTS=-Xms2g -Xmx2g" -e "OPENSEARCH_PATH_CONF=/usr
...
+```
+{% include copy.html %}
+
+A variable set to an empty value is skipped rather than passed through, so it cannot be used to clear a value set in `opensearch.yml`.
+
+## Setting precedence
+
+A setting can be supplied in more than one place. When it is, OpenSearch uses the first value it finds in the following order:
+
+1. Transient cluster settings, applied with the Cluster Settings API.
```
reta, line 108: I think we are missing:
 - JVM system properties
 - Process environment variables

_install-and-configure/configuring-opensearch/index.md
```diff
@@ -77,6 +92,27 @@ docker run -e "OPENSEARCH_JAVA_OPTS=-Xms2g -Xmx2g" -e "OPENSEARCH_PATH_CONF=/usr
...
+2. Persistent cluster settings, applied with the Cluster Settings API.
+3. Settings passed as `-E` flags at startup.
+4. Settings in `opensearch.yml`.
+5. The default value of the setting.
+
+Startup flags take precedence over the configuration file because `opensearch.yml` is read first and the `-E` values are applied on top of it. A `${VAR}` placeholder in `opensearch.yml` is resolved from the environment as the file is read, so a value supplied in that way behaves as though it had been written into the file.
```
reta, line 114:
```suggestion
Startup flags take precedence over the configuration file because `opensearch.yml` is read first and the values specified with `-E` flag are applied on top of it. A `${ENV_VAR}` placeholder in `opensearch.yml` is resolved from the environment and substituted with its value.
```

reta:
Thanks @gohri81-tech , a few minor comments, looks good otherwise

_install-and-configure/configuring-opensearch/index.md
```diff
@@ -77,6 +92,27 @@ docker run -e "OPENSEARCH_JAVA_OPTS=-Xms2g -Xmx2g" -e "OPENSEARCH_PATH_CONF=/usr
...
 
 
+Docker is the exception to the dot restriction described in [Directly in the shell environment](#directly-in-the-shell-environment). The image's entrypoint reads the container environment and converts any variable whose name looks like a setting, meaning at least two dot-separated lowercase words, plus `processors`, into an `-E` flag:
+
+```bash
+docker run -e "discovery.type=single-node" -e "cluster.name=my-cluster" opensearchproject/opensearch:latest
+```
+{% include copy.html %}
+
+A variable set to an empty value is skipped rather than passed through, so it cannot be used to clear a value set in `opensearch.yml`.
```
reta, line 102: Special case here if you want to mention, empty array like: `-E node.roles=[]`

[1] https://github.com/opensearch-project/OpenSearch/pull/10625

_install-and-configure/configuring-opensearch/index.md
gohri81-tech: On process environment variables: `prepareEnvironment` loads `opensearch.yml` and then applies the `-E` properties on top of it, and `Settings.Builder.replacePropertyPlaceholders()` resolves `${ENV_VAR}` from `System::getenv` while the file is being read. So an environment variable takes effect at the same level as the file rather than at one of its own, which is what the paragraph under the list now says. Happy to give it its own entry if you would rather see it listed.

On JVM system properties: I could not find a path on main that reads them as settings - `initializeSettings` only puts the input settings and the command line properties - so I left them out rather than guess. If you mean `-D` values passed through `OPENSEARCH_JAVA_OPTS`, or somewhere else that picks them up, point me at it and I will add both entries in one go.

_install-and-configure/configuring-opensearch/index.md
gohri81-tech: Added in the latest commit: the paragraph now says that `[]` is not an empty value, so `-e "node.roles=[]"` is turned into `-Enode.roles=[]` and does set the roles to an empty list, with a link to opensearch-project/OpenSearch#10625. Reword it if you would put it differently.

_install-and-configure/configuring-opensearch/index.md
reta: Thanks @gohri81-tech , would be good to have env variable included exactly since they will be used if there are no overrides from the command line (-E flag). Regarding System properties, I think you are right, we do not take those into account during substitution - no need to mention (apologies for confusion)

_install-and-configure/configuring-opensearch/index.md
```diff
@@ -77,6 +92,27 @@ docker run -e "OPENSEARCH_JAVA_OPTS=-Xms2g -Xmx2g" -e "OPENSEARCH_PATH_CONF=/usr
...
 
 
+Docker is the exception to the dot restriction described in [Directly in the shell environment](#directly-in-the-shell-environment). The image's entrypoint reads the container environment and converts any variable whose name looks like a setting, meaning at least two dot-separated lowercase words, plus `processors`, into an `-E` flag:
+
+```bash
+docker run -e "discovery.type=single-node" -e "cluster.name=my-cluster" opensearchproject/opensearch:latest
+```
+{% include copy.html %}
+
+A variable set to an empty value is skipped rather than passed through, so it cannot be used to clear a value set in `opensearch.yml`. An empty list is the exception: `[]` is not an empty value, so `-e "node.roles=[]"` is turned into `-Enode.roles=[]` and does set the roles to an empty list, as suggested in [opensearch-project/OpenSearch#10625](https://github.com/opensearch-project/OpenSearch/issues/10625).
```
reta:
```suggestion
A variable set to an empty value is skipped rather than passed through, so it cannot be used to clear a value set in `opensearch.yml`. A special `[]` value denotes empty list notation where applicable.
```

